PR #27, opened 2019-05-03
GCP-guest: use non-wrapper versions when decorated

hopkiw:
`decorate:true` does seem to do the work our wrapper does, so testing with some non-wrapper versions of these images.


wrigri:
LGTM

PR #48, opened 2019-06-11
Update image for cpp builds

hopkiw:
`debian` doesn't have the necessary build tools and this is a make based repo, don't want to clutter it with shell scripts specific to prow or have a super customized container image for this either. compromised on debian+build-essential as an available image.


hopkiw:
/uncc subratp
should update OWNERS

PR #87, opened 2019-09-16
Testgrid: Add Automatic Image Deployment

chases2:
Automatically pushes images when a change is made to a testgrid command

Hold for secrets to be added
/hold

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
@@ -2,7 +2,7 @@ presubmits:
   GoogleCloudPlatform/testgrid:
   - name: test-testgrid-all
     branches:
-    - master
+    - ^master$
     decorate: true
     always_run: true
     spec:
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
```diff
@@ -14,3 +14,20 @@ presubmits:
...
+    decorate: true
+    run_if_changed: '^cmd/.*$'
+    spec:
+      containers:
+      - image: launcher.gcr.io/google/bazel:0.29.1
+        command:
+        - bazel
+        args:
+        - run
+        - //images:push
```
fejta: I would expect this to need credentials

michelle192837:
Also if you locally tested this to verify, note in the PR description.

chases2:
@michelle192837 Still working on getting this to test correctly locally; will remove the hold once I have that verification

prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
fejta: I suspect you'll need to activate the service account credentials too? So might be best to use something like bazel.sh -- and have this activate the service account if the env var is set.

https://github.com/kubernetes/test-infra/blob/master/hack/bazel.sh

https://github.com/kubernetes/test-infra/blob/3dfd16c0c1574a52764747a394e8ec2594796239/boskos/update_prow_config.sh#L42-L44

chases2:
I tested this change locally, but it's now dependent on a script in the repository; see https://github.com/GoogleCloudPlatform/testgrid/pull/14

prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
```diff
@@ -13,3 +13,26 @@ presubmits:
...
         - test
         - //...
+
+postsubmits:
+  chases2/testgrid:
+  - name: push-testgrid-images
+    branches:
+    - ^master$
+    decorate: true
+    run_if_changed: '^cmd/.*$'
```
fejta: This is not broad enough. If I change something in proto/ or metadata/ we'll want a new image with this change.

chases2: It's not just the various protos or metadata; it's most of the repository.
I've changed this regex to reflect this.

chases2:
Fix: changed job from local branch to base branch

PR #96, opened 2019-09-25
Setup a general build cluster for oss-prow

clarketm:
Summary of changes:
1. Pulled in the `Makefile.gcloud.mk` file, used internally and in Istio  which contains a bunch of helpers.
2. Separated out `grandmatriarch` *default* and *test-pods* deployments into separate files.
3. Added `deploy-build` command to `Makefile` for deploying build cluster.
4. Enabled `build` cluster on test-infra presubmits.

/assign @cjwagner @fejta

### DIFF
--- a/prow/Makefile
+++ b/prow/Makefile
@@ -12,24 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+include Makefile.gcloud.mk
+
 # These are the usual GKE variables.
 PROJECT       ?= oss-prow
-BUILD_PROJECT ?= oss-prow
+PROJECT_BUILD ?= oss-prow-builds
 ZONE          ?= us-west1-a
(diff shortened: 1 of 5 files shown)


clarketm:
As soon as this merges, I am going to follow-up with a Prow auto-deploy postsubmit.

prow/Makefile
```diff
@@ -12,25 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+include Makefile.gcloud.mk
```
fejta: Since this isn't istio, can we use bazel rules and avoid Makefiles?

clarketm: Yes, we can do that. I created this #97. The Makefile is simple enough so that it shouldn't be too much effort to port.

Issue #97, opened 2019-09-26
Replace Make with Bazel


clarketm:
Currently, this repo uses Makefiles to perform *build* and *deploy* operations. We would like to replace each `make` **target** with its equivalent `bazel` command.

/help

Issue #94, opened 2019-09-24
Create a dedicate build cluster


clarketm:
### Create a dedicate build cluster
This allows for isolation between the Prow cluster and the jobs running on it. Furthermore, this enables the introduction of "trusted" jobs for general deployments, image builds, and prow bumps in order to automate repeatable, mundane tasks without exposing the cluster (or secrets) to unprivileged users.

- [x] Create a GCP Prow **build** cluster.
- [x] Migrate jobs to the **build** cluster.
- [x] Set the **build** cluster as the *default* cluster.

/assign

PR #99, opened 2019-09-26
Setup Prow auto-deploy

clarketm:
Summary of changes:
1. Move deployments/services to `cluster` directory.
2. Create a *"trusted"* prow deploy job.

/assign @cjwagner @krzyzacy @fejta 

> Once this goes through, I will set the `build` cluster as *default* #94 and migrate jobs to it. Also, update this to use bazel #97.


prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -30,3 +28,25 @@ presubmits:
...
+      spec:
+          containers:
+          - image: gcr.io/k8s-testimages/gcloud-in-go:v20190125-cc5d6ecff3
+            command:
+            - make
+            args:
+            - -C
+            - prow
+            - deploy
+            - deploy-build
```
cjwagner: Is it safe to use both these targets at the same time? If make builds all the target dependencies first or in parallel then we could unintentionally deploy all the resources to one of the clusters.

clarketm: Make, by *default*, runs each target in a separate subshell ([ref](https://www.gnu.org/software/make/manual/make.html#Execution)). Also, as a additional layer of security in general, we are avoiding updating of *global* state (i.e. `KUBECONFIG`) by using a temp file for each execution.

https://github.com/GoogleCloudPlatform/oss-test-infra/blob/d15f44bc53127d4dc3af73fb0d2abb75f7981d9a/prow/Makefile.gcloud.mk#L24-L27

clarketm: > We should probably move all the "non-trusted" jobs out of the cluster before adding the deployer credentials to it.

In the fine print I mentioned as soon as this merges, I am going to move all "non-trusted" jobs to the `build` cluster (https://github.com/GoogleCloudPlatform/oss-test-infra/pull/99#issue-321850303). Although it may make m ore sense to do it before.

/hold

clarketm: I have migrated "non-trusted" jobs to the `build` cluster and uploaded the `prow-deployer` secret.

/hold cancel

clarketm:
@cjwagner - I made the switch. The `build` cluster is now the *default*, so no jobs are running on the *trusted* cluster.

PR #100, opened 2019-09-26
Enable OSS rerun for all googlers

### LINKED ISSUE #68: Setup Prow job rerun
image

It would be great to expose this feature to existing prow users. The infrastructure is in place - it just needs to be configured.

Tasks:
- [x] A `GoogleCloudPlatform` owner to create a [Github Oauth App](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/deck/github_oauth_setup.md#set-up-secrets) (step 1 and 2)
- [x]  Define which *users* or *teams* will be delegated rerun permission.
- [x] Update the `deck` Kubernetes deployment and secrets (`github-oauth-config` and `cookie`)., #100

clarketm:
Enable `deck` rerun for [Googlers](https://github.com/orgs/GoogleCloudPlatform/teams/googlers).
closes #68

> `github-oauth-config` and `cookie` secrets loaded to cluster.

/assign @cjwagner @krzyzacy

### DIFF
--- a/prow/config.yaml
+++ b/prow/config.yaml
@@ -84,6 +84,10 @@ deck:
         required_files:
           - artifacts/junit.*\.xml
     announcement: "The old job viewer, Gubernator, has been deprecated in favour of this page, Spyglass.{{if .ArtifactPath}} For now, the old page is <a href='https://gubernator-internal.googleplex.com/build/{{.ArtifactPath}}'>still available</a>.{{end}} Please send feedback to osp-engprod@google.com."
+  rerun_auth_config:
+    github_team_slugs:
+    - org: GoogleCloudPlatform
+      slug: googlers
 
 presets:
 - labels:
--- a/prow/cluster/cluster.yaml
+++ b/prow/cluster/cluster.yaml
@@ -188,11 +188,23 @@ spec:
         - --tide-url=http://tide/
         - --hook-url=http://hook:8888/plugin-help
         - --redirect-http-to=prow.gflocks.com
-        - --github-oauth-config-file=/etc/github/oauth
+        - --rerun-creates-job
+        - --oauth-url=/github-login
+        - --github-token-path=/etc/github/oauth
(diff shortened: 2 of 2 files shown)


prow/config.yaml
```diff
@@ -83,3 +83,7 @@ deck:
         required_files:
           - artifacts/junit.*\.xml
     announcement: "The old job viewer, Gubernator, has been deprecated in favour of this page, Spyglass.{{if .ArtifactPath}} For now, the old page is <a href='https://gubernator-internal.googleplex.com/build/{{.ArtifactPath}}'>still available</a>.{{end}} Please send feedback to osp-engprod@google.com."
+    rerun_auth_config:
+      github_team_slugs:
+      - org: GoogleCloudPlatform
+        slug: googlers
```
cjwagner: We'll probably want to set up ghproxy for this cluster since listing the member of this team currently costs 25 API ratelimit tokens. That being said, the cache entry would become invalid pretty quickly as I suspect that team has high churn and this bot token isn't heavily utilized.

cjwagner:
Did you add the two new secrets?

clarketm:
@cjwagner yes, the two secrets are added to the cluster. 
https://github.com/GoogleCloudPlatform/oss-test-infra/pull/100#issue-321924132

prow/config.yaml
clarketm: Good idea. Adding `ghproxy` can only aid in minimizing token usage, even if the cache entry does get invalidated. I can add a ghproxy deployment for this instance, #101.

fejta:
Should we block this on getting ghproxy setup?

clarketm:
@fejta 
I can create the `ghproxy` depoyment (in code) now and PR. Not sure if this should be blocking or not.

prow/config.yaml
clarketm: ghproxy will be added in #128

Issue #126, opened 2019-10-10
We do not validate that plugin.yaml, config.yaml is valid before merge


fejta:
See https://github.com/GoogleCloudPlatform/oss-test-infra/pull/125

fejta:
/assign @cjwagner 

Can you look into this? It sounded like you know what's going on from yesterday

PR #131, opened 2019-10-11
Fix prow config validation.

cjwagner:
fixes https://github.com/GoogleCloudPlatform/oss-test-infra/issues/126
/assign @fejta

### DIFF
--- a/prow/plugins.yaml
+++ b/prow/plugins.yaml
@@ -48,6 +48,7 @@ plugins:
   - shrug
   - size
   - trigger
+  - verify-owners
   - yuks
 
   GoogleCloudPlatform/oss-test-infra:
@@ -75,5 +76,13 @@ plugins:
   - pony # /pony
   - shrug # /shrug
   - size # auto-add size label
+  - verify-owners # validate OWNERS file changes
   - wip # draft-PR awareness
   - yuks # /joke
+
+  # This is a gerrit repo so this config doesn't do anything. It is included
+  # to satisfy the checkconfig tool which expects all repos that configure
+  # jobs to enable the trigger plugin if a plugins.yaml is specified.
+  # https://github.com/kubernetes/test-infra/issues/14743
+  https://kunit-review.googlesource.com/linux:
+  - trigger
--- a/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
@@ -15,6 +15,11 @@ presubmits:
             args:
               - --config-path=./prow/config.yaml
               - --job-config-path=./prow/prowjobs
+              - --plugin-config=prow/plugins.yaml
+              - --strict
+              # This warning can forbid valid (and convenient) config. Exclude it.
+              # Details: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/131#discussion_r334208082
+              - --exclude-warning=mismatched-tide
 
     - name: pull-test-infra-go-test
       branches:


cjwagner:
Output when running the updated presubmit against this PR with pj-on-kind.sh:
```
{"component":"checkconfig","file":"prow/config/config.go:930","func":"k8s.io/test-infra/prow/config.(*Config).finalizeJobConfig","level":"warning","msg":"default_decoration_config will be deprecated on April 2020, and it will be replaced with default_decoration_configs['*'].","time":"2019-10-11T23:50:27Z"}
{"component":"checkconfig","file":"prow/cmd/checkconfig/main.go:288","func":"main.main","level":"info","msg":"checkconfig passes without any error!","time":"2019-10-11T23:50:27Z"}
```

BenTheElder:
another bug: spiffxp is marked busy / on vacation on github but still being assigned, is that something we have to enable for this repo?

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -5,7 +5,7 @@ presubmits:
       branches:
         - master
       decorate: true
-      always_run: true
+      run_if_changed: '^prow/(cluster/|prowjobs/|config.yaml$|plugins.yaml$)'
```
BenTheElder: this job seems pretty cheap and most of this repo is yaml? 🤷‍♂ 
I continue to claim `run_if_changed` is a footgun to let changes slip through.

cjwagner: /shrug
Switched back to `always_run: true`.

prow/plugins.yaml
```diff
@@ -75,5 +76,13 @@ plugins:
...
   - size # auto-add size label
+  - verify-owners # validate OWNERS file changes
   - wip # draft-PR awareness
   - yuks # /joke
+
+  # This is a gerrit repo so this config doesn't do anything. It is included
+  # to satisfy the checkconfig tool which expects all repos that configure
+  # jobs to enable the trigger plugin if a plugins.yaml is specified.
+  # https://github.com/kubernetes/test-infra/issues/14743
+  https://kunit-review.googlesource.com/linux:
```
BenTheElder: xref: https://github.com/kubernetes/test-infra/issues/14743
(commenting so when we fix that issue we can easily find this change)

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -15,6 +15,9 @@ presubmits:
             args:
               - --config-path=./prow/config.yaml
               - --job-config-path=./prow/prowjobs
+              - --plugin-config=prow/plugins.yaml
+              - --strict
+              - --exclude-warning=mismatched-tide
```
BenTheElder: what does this do?

cjwagner: This part is a little weird. There are two tide mismatch warnings, the one we are excluding and the lenient one we are still running: 
https://github.com/kubernetes/test-infra/blob/191b281468351f16bc560e55ee17b4a590db433c/prow/cmd/checkconfig/main.go#L88-L89

The `mistmatched-tide-lenient` warning ensures that all repos that configure Tide to require a label that is associated with a plugin also enable that plugin. This catches the case where you require `lgtm` to merge, but forget to enable the `lgtm` plugin.

The `mismatched-tide` warning **also** checks that all repos that configure Tide to forbid a label that is associated with a plugin also enable that plugin. This catches the case where you forbid `do-not-merge/work-in-progress`, but for get to enable the `wip` plugin.

I don't really like this check because we often intentionally make a Tide query cover multiple repos where one of the repos enables a plugin that manages a blocker label, but the other repos don't. Tide behaves properly because there are never any PRs with the forbidden label in the repo where the blocker plugin is not enabled and this pattern is pretty nice for keeping config readable so I don't consider it a misconfiguration and don't want checkconfig to ban it.

BenTheElder: interesting, thanks!

I wonder if it might be better to use an explicit lint list or adjust the defaults, golang-ci lint has some flags that help with this (`--disable-all`)

might also be worth a shorter comment in the config file

BenTheElder:
/lgtm
/hold
might be worth commenting the config a little

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
cjwagner: We support either a whitelist or blacklist for specifying which warnings to run: https://github.com/kubernetes/test-infra/blob/191b281468351f16bc560e55ee17b4a590db433c/prow/cmd/checkconfig/main.go#L152-L153
An explicit lint list would be less prone to breaking, but I worry that we'd add new important checks as we add new component/functionality and forget to update the job spec. If there are backwards incompatible checks that are added, we can disable them with a 1 line fix.

Either way seems fine to me though. Let me know if you want me to switch to an explicit list.

BenTheElder: i'm not actually sure what is better, it seems like ideally you wouldn't have disagreement about the defaults though!

BenTheElder: (and if you do, maybe we should explicitly opt-in to lints)

BenTheElder:
/lgtm
/approve
/hold cancel
thanks!

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
cjwagner: Its just that one warning and I'd like to avoid having to explicitly opting in. 
@stevekuznetsov what do you think about making that check default to disabled? It would be nice to default to running checks that are appropriate for all Prow instances and then opt-in to any additional ones.

stevekuznetsov: I like the check :)

stevekuznetsov: I do agree that making one of the two default to off would be nice.

cjwagner: I understand why some people might want to enable the check, but I think we should make the default behavior of `checkconfig` appropriate for any Prow instance. There are times when it makes sense and should be valid to provide configuration that violates the `mismatched-tide` check. It can be more efficient and easier to read, especially when plugins cannot be configured to have no-op behavior (e.g. [`mergecommitblocker`](https://github.com/kubernetes/test-infra/tree/master/prow/plugins/mergecommitblocker)).
Previous [discussion](https://github.com/kubernetes/test-infra/pull/9395#issuecomment-429483447).

PR #149, opened 2019-10-28
Move push-testgrid-images to trusted cluster

chases2:
The necessary secret is present in the trusted cluster

In response to https://github.com/GoogleCloudPlatform/testgrid/issues/43

/assign @Katharine 
/cc @fejta @michelle192837


Katharine:
You'll need to change either the location of this job or the test to let this merge. If you change the test, you will also need to ensure suitable ownership for this file - in general we don't want almost anyone to be able to make changes to trusted jobs.

chases2:
Update: moved prow job to `gcp-oss-test-infra-config.yaml` to keep trusted jobs together

Issue #152, opened 2019-10-29
Add support for testgrid annotations on prowjobs


fejta:
Can we get your utility set up on this repo to a) automatically create/update a testgrid.yaml based on the annotations on prowjobs in this repo and b) create kubernetes/test-infra PRs to get this added to testgrid.k8s.io?

/assign @chases2

chases2:
Yes. I plan to start this once the Transfigure image is working as intended for Istio.

Then, I can create docs for an example Prow Job and use it here.

chases2:
Tests are currently running, but aren't passing. See https://github.com/GoogleCloudPlatform/oss-test-infra/pull/189

chases2:
Tests are now passing. See #193

PR #158, opened 2019-11-07
Add Prow Job to push TestGrid Configs to k8s-testgrid

chases2:
This should enable Prow annotations, and allow TestGrid to automatically update k8s/test-infra as other sources are.

/assign @fejta @michelle192837 
/cc @hopkiw

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
@@ -53,6 +53,33 @@ postsubmits:
         - prow
         - deploy
         - deploy-build
+  - name: post-test-infra-upload-testgrid-config
+    cluster: test-infra-trusted
+    run_if_changed: '^(prow/prowjobs/.*\.yaml)|(testgrid/config\.yaml)$'
+    decorate: true
+    branches:
+    - master
+    annotations:
+      testgrid-create-test-group: "false"
+    spec:
+      containers:
+      - image: gcr.io/k8s-prow/transfigure
+        command:
+        - /transfigure.sh
+        args:
+        - /etc/github-token/oauth
+        - prow/config.yaml
+        - prow/prowjobs
+        - testgrid/config.yaml
+        - GoogleCloudPlatform
+        volumeMounts:
+        - name: github
+          mountPath: /etc/github-token
+          readOnly: true
+      volumes:
+      - name: github
+        secret:
+          secretName: oauth-token
   GoogleCloudPlatform/testgrid:
   - name: push-testgrid-images
     cluster: test-infra-trusted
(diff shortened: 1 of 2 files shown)


testgrid/config.yaml
```diff
@@ -0,0 +1,5 @@
+# Dashboards need to be specified here to be created on TestGrid
+# A prow annotation will be invalid if it references a dashboard that doesn't exist
+dashboards:
```
fejta: Do we need a) this file and b) these keys in it?

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -53,6 +53,33 @@ postsubmits:
...
         - deploy
         - deploy-build
+  - name: post-test-infra-upload-testgrid-config
+    cluster: test-infra-trusted
+    run_if_changed: '^(prow/prowjobs/.*\.yaml)|(testgrid/config\.yaml)$'
+    decorate: true
+    branches:
+    - master
+    annotations:
+      testgrid-create-test-group: "false"
```
fejta: Why not?

testgrid/config.yaml
chases2: We do need this file, as Configurator needs a base config.yaml file to append Prow Annotations to. Configurator doesn't create TestGrid YAML out of nothing.

As the comment says, dashboards will need to be enumerated here. A shell file is needed anyway, so why not add the correct keys and a description for what will go here?

fejta: IMO if it works with this file and fails without it, then (long-term) we should remove that constraint because its pretty ridiculous. There's nothing meaningful in this file.

Long term, do we really need this file? Why can't I annotate my jobs with dashboard, dashboard group, tab options, etc and have it generate everything?

chases2:
Addresses #152

fejta:
@BenTheElder @chases2 what's with the two pending jobs here? https://prow.gflocks.com/

testgrid/config.yaml
chases2: The big advantage to having the list of dashboards explicitly like this is that Configurator can point out (and fail) when you misspell a dashboard. If there's no "source of truth" for that, then there's no way for Configurator to know whether `sig-tset-infra-king` is incorrect, or whether that's my brand new dashboard.

I think it's more valuable to have a singular, though separate, list of dashboards and have annotations tie them to them.

chases2:
Don't know yet; tracking at #159

testgrid/config.yaml
fejta: That doesn't seem very convincing to me.

a) we can still validate that dashboards all start with sig- or google- like we do today
b) nothing preventing someone from fat fingering things in this yaml file
c) it is trivial to send a follow up PR to change sig-testing-king to sig-testing-kind

The test group and its thousands of GCS files are the only thing that's consequential to rename (aka if you rename the testgroup path you lose test history)

PR #161, opened 2019-11-08
Addition of TestGrid dashboards

chases2:
Additions will be necessary to remove conflicts and add tests so that configuration will pass in correctly.

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
@@ -35,12 +35,14 @@ presubmits:
 
 postsubmits:
   GoogleCloudPlatform/oss-test-infra:
-  - name: post-test-infra-deploy-prow
+  - name: post-oss-test-infra-deploy-prow
     cluster: test-infra-trusted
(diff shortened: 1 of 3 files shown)


chases2:
xref #152

chases2:
@fejta @hopkiw This change includes some decisions on how these jobs are logged in TestGrid; please review and let me know if this looks good. Thanks!

testgrid/config.yaml
```diff
@@ -1,5 +1,7 @@
 # Dashboards need to be specified here to be created on TestGrid
 # A prow annotation will be invalid if it references a dashboard that doesn't exist
 dashboards:
+  - name: google-gcp-guest
```
fejta: Style is generally to keep list items indented at the same level as the key, thus:
```yaml
dashboards:
- name: foo
- name: bar
```

chases2: Done

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -86,6 +88,9 @@ postsubmits:
     branches:
     - ^master$
     decorate: true
+    annotations:
+      testgrid-dashboards: google-oss-infra
+      testgrid-alert-email: slchase@google.com
```
fejta: Eventually the oncall list please!

PR #224, opened 2020-01-10
Annotate grandmatriarch service-account

fejta:
ref #202 

This will allow me to bind the service accounts together and send a PR to switch it over to using workload-identity

Issue #202, opened 2020-01-07
Use workload identity for prow


fejta:
We should stop sending this flag: 
https://github.com/GoogleCloudPlatform/oss-test-infra/blob/bc1c6b91cf772fa7f116a3c2119a874e53f2c19b/prow/cluster/cluster.yaml#L344

We already define a service account:
https://github.com/GoogleCloudPlatform/oss-test-infra/blob/bc1c6b91cf772fa7f116a3c2119a874e53f2c19b/prow/cluster/cluster.yaml#L335

We should annotate it with GCP SA rights: https://github.com/kubernetes/test-infra/tree/master/experiment/workload-identity

If the flag is unset then it will use default creds:
https://github.com/kubernetes/test-infra/blob/master/prow/cmd/gerrit/main.go#L240

https://github.com/kubernetes/test-infra/blob/164c5e85105f85bb8fd8c181a085911a3d1010fd/pkg/io/opener.go#L57-L60

https://godoc.org/cloud.google.com/go/storage#hdr-Creating_a_Client

fejta:
Merged #204 

Binding:
```console
fejta@fejta3:~/src/gh/test-infra$ experiment/workload-identity/bind-service-accounts.sh oss-prow us-west1-a prow default gerrit oss-prow@oss-prow.iam.gserviceaccount.com
+ gcloud iam service-accounts add-iam-policy-binding --project=oss-prow --role=roles/iam.workloadIdentityUser '--member=serviceAccount:oss-prow.svc.id.goog[default/gerrit]' oss-prow@oss-prow.iam.gserviceaccount.com
Updated IAM policy for serviceAccount [oss-prow@oss-prow.iam.gserviceaccount.com].
```

fejta:
Well that failed miserably, apparently because I cannot follow instructions: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_workload_identity_on_an_existing_cluster

fejta:
https://github.com/GoogleCloudPlatform/oss-test-infra/blob/4a6428605cba8c4053d0371bca9f2bf82ece031a/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml#L114-L116

now uses 

[...]

PR #226, opened 2020-01-13
Use workload-identity for grandmatriarch

fejta:
ref #202 

https://github.com/GoogleCloudPlatform/oss-test-infra/pull/224#issuecomment-573816169 validates that all three service accounts can access the GCP service account


fejta:
```console
secret/http-cookiefile replaced
successfully updated token, sleeping for 20m...
```

PR #240, opened 2020-01-16
switch tide to clusterrole so it can create pods in other namepsaces

BenTheElder:
... also move to v1

/cc @fejta

### DIFF
--- a/prow/cluster/cluster.yaml
+++ b/prow/cluster/cluster.yaml
@@ -868,10 +868,10 @@ metadata:
   namespace: default
   name: "tide"
 ---
-kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  namespace: default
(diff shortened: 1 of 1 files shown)


prow/cluster/cluster.yaml
```diff
@@ -868,10 +868,10 @@ metadata:
   namespace: default
   name: "tide"
 ---
-kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
```
fejta: All prow jobs live in a single namespace (the same namespace as where tide lives). So I don't understand why this is necessary?

BenTheElder: the logs show tide failing to create pods in test-pods.
we're using clusterrole for our other components :man_shrugging:

BenTheElder: we could create roles for each namespace instead but i'm not sure what we intend to target.

fejta: Tide doesn't create pods?

BenTheElder: ACK, logs read wrong, discussing in chat/

PR #268, opened 2020-02-12
Use image to choose bazel version in testgrid repo

fejta:
/assign @michelle192837 @chases2 

Will unblock https://github.com/GoogleCloudPlatform/testgrid/pull/92

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
@@ -7,7 +7,10 @@ presubmits:
     always_run: true
     spec:
       containers:
-      - image: launcher.gcr.io/google/bazel:2.0.0
+      - image: launcher.gcr.io/google/bazel:2.1.0
+        env:
+        - name: USE_BAZEL_VERSION
+          value: real # Ignore .bazelversion in CI
         command:
         - bazel
         args:


prow/prowjobs/GoogleCloudPlatform/testgrid/testgrid-images.yaml
```diff
@@ -7,7 +7,10 @@ presubmits:
     always_run: true
     spec:
       containers:
-      - image: launcher.gcr.io/google/bazel:2.0.0
+      - image: launcher.gcr.io/google/bazel:2.1.0
+        env:
+        - name: USE_BAZEL_VERSION
```
chases2: What does this environment variables do? With or without it, doesn't this use version 2.1.0?

There is no `.bazelversion` file in this repository.

fejta: See https://prow.gflocks.com/view/gcs/oss-prow/pr-logs/pull/GoogleCloudPlatform_testgrid/92/test-testgrid-all/1227394014076997632 for what happens without this

fejta: It blows up because `.bazelversion says 2.1.0` when the image has `2.0.0` (or whatever), despite the fact that it will work with either.

PR #271, opened 2020-02-13
add space for espv2 jobs

TAOXUY:


prow/prowjobs/GoogleCloudPlatform/esp-v2/esp-v2.yaml
```diff
@@ -20,6 +20,9 @@ presubmits:
       - name: cloudesf-testing-github-prow-service-account
         secret:
           secretName: cloudesf-testing-github-prow-service-account
+      - name: espv2-disk
+        persistentVolumeClaim:
```
fejta: You'll need to want to mount this volume somewhere and make sure the job is using it (like you're doing with the `cloudesf-testing-github-prow-service-account` volume)

prow/cluster/espv2-disk.yaml
```diff
@@ -0,0 +1,11 @@
+kind: PersistentVolumeClaim
```
fejta: This needs to be part of the build cluster (where jobs run) not the service cluster (prow's control plane)

TAOXUY: You mean this line? or this whole config

prow/pvsc/GoogleCloudPlatform_esp-v2_pvsc.yaml
```diff
@@ -0,0 +1,11 @@
+kind: PersistentVolumeClaim
```
fejta: a) PVCs (c and s are transposed)
b) prow/build/pvcs (this is for the build cluster)
c) Update the rule to apply this: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/29de34218ad9075ee82feabc558b03b20bb26960/prow/Makefile#L37-L40

prow/cluster/espv2-disk.yaml
fejta: The file

PR #278, opened 2020-02-15
Delete gcp-esp-v2-default serviceaccount from shared cluster

fejta:
The espv2 jobs no longer run in the shared cluster, so we no longer need this service account

/assign @TAOXUY


TAOXUY:
LGTM

PR #288, opened 2020-02-20
Enable reporting of pod info to GCS.

michelle192837:
Also tweaks things like pod timeout to match kubernetes/test-infra for
more useful debugging.

Lens will be enabled in follow-up PR if this works.


Katharine:
Does this need some accompanying service account change so that crier can actually write to GCS?

(It should have exactly the permissions that is in the secret podutils uses)

michelle192837:
Mmk, I believe the crier service account is now updated, and the change to actually add the lens is added (assuming that it's fine to add both at the same time).

PR #314, opened 2020-03-07
Add presubmit and continuous coverage jobs

nareddyt:


nareddyt:
Will be unblocked when #315 is merged

PR #267, opened 2020-02-11
Mandate TestGrid Validation

chases2:
Also makes changes to ensure that this presubmit passes:
- Dashboard 'googleoss-esp-v2' removed; no postsumits/periodics are run by that team

PR #315, opened 2020-03-09
Add esp-v2 to testgrid

nareddyt:
Otherwise, we cannot modify our jobs due to the validations added in #267

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/esp-v2/esp-v2.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/esp-v2/esp-v2.yaml
@@ -509,6 +509,8 @@ periodics:
     decorate: true
     decoration_config:
         timeout: 5h
+    annotations:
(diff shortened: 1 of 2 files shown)


prow/prowjobs/GoogleCloudPlatform/esp-v2/esp-v2.yaml
```diff
@@ -25,6 +27,8 @@ presubmits:
     cluster: espv2
     always_run: true
     decorate: true
+    annotations:
+      testgrid-dashboards: googleoss-test-infra
```
chases2: I think these jobs should go on the new `googleoss-esp-v2` dashboard you created, right? They're in the espv2 cluster.

nareddyt: Oops, good catch

PR #321, opened 2020-03-11
Domain is now oss-prow.knative.dev

chases2:
Changes domain name in all documentation

Does not cover domain name in [prow/cluster/cluster.yaml](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/a9beebb04e4e85a178c994226a1408ef8680401c/prow/cluster/cluster.yaml#L306-L313); I'm not sure if that managed certificate is being used currently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,6 @@ limitations under the License.
 
 Here lives the OSS [prow](https://github.com/kubernetes/test-infra/tree/master/prow) config for google-owned OSS projects.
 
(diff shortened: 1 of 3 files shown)


spiffxp:
I don't have access to see if the cert is being used, but I would recommend removing it

chases2:
That change might actually cause a problem somewhere, so I put it in a different PR. See #323

PR #324, opened 2020-03-13
New deck URL is oss-prow.knative.dev

fejta:


chases2:
This appears to be a subset of #321

PR #336, opened 2020-04-14
Move cAdvisor prow config from k8s.io

dashpole:
Let me know if there is anything else required.  I haven't checked that this works, so it may need further changes.

https://github.com/kubernetes/test-infra/issues/12863

/assign @krzyzacy

### DIFF
--- a/prow/prowjobs/google/cadvisor/OWNERS
+++ b/prow/prowjobs/google/cadvisor/OWNERS
@@ -1,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- dashpole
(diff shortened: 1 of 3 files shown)


prow/prowjobs/google/cadvisor/OWNERS
```diff
@@ -0,0 +1,4 @@
+# See the OWNERS docs at https://go.k8s.io/owners
+
+approvers:
+- dashpole
```
fejta, line 4: Have you joined GCP org yet?

dashpole:
cc @bobbypage @SergeyKanzhelev @dims

I'm not going to be able to complete this.  Depending on whether or not cAdvisor is moved somewhere else, this may or may not be neccessary.

PR #344, opened 2020-04-21
Deploy and use ghproxy for Prow components.

cjwagner:
/assign @fejta

### DIFF
--- a/prow/cluster/ghproxy.yaml
+++ b/prow/cluster/ghproxy.yaml
@@ -1,0 +1,63 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  namespace: default
+  labels:
+    app: ghproxy
+  name: ghproxy
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 50Gi
+---
+apiVersion: v1
+kind: Deployment
+metadata:
+  namespace: default
+  name: ghproxy
+  labels:
+    app: ghproxy
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: ghproxy
+    spec:
+      containers:
+      - name: ghproxy
+        image: gcr.io/k8s-prow/ghproxy:v20200417-a6017ee5b
+        args:
+        - --cache-dir=/cache
+        - --cache-sizeGB=49
+        - --serve-metrics=true
+        ports:
+        - containerPort: 8888
+        - containerPort: 9090
+        volumeMounts:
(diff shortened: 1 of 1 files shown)


prow/cluster/ghproxy.yaml
```diff
@@ -0,0 +1,63 @@
...
+    app: ghproxy
+  name: ghproxy
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 50Gi
+---
+apiVersion: extensions/v1beta1
```
fejta, line 15: Deployments are v1 now, can you switch that?

prow/cluster/cluster.yaml
```diff
@@ -483,6 +487,8 @@ spec:
         - --config-path=/etc/config/config.yaml
         - --job-config-path=/etc/job-config
         - --dry-run=false
+        - --github-endpoint=http://ghproxy
```
fejta, line 490: Can we first deploy it and then make things use it?

cjwagner: Sure, I'll split it up. I was going to do separate PRs originally, but the GH client will fallback to using the API directly if ghproxy isn't working for any reason so I didn't think it was very risky.

prow/cluster/ghproxy.yaml
cjwagner: Done.

prow/cluster/cluster.yaml
fejta: In that case might be fine to just do it but either way

cjwagner: Done.
Follow up: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/345

prow/cluster/ghproxy.yaml
```diff
@@ -0,0 +1,63 @@
...
+        - containerPort: 9090
+        volumeMounts:
+        - name: cache
+          mountPath: /cache
+      volumes:
+      - name: cache
+        persistentVolumeClaim:
+          claimName: ghproxy
+---
+apiVersion: v1
```
fejta, line 47: apps/v1

https://github.com/kubernetes/test-infra/blob/e0b6b08e020b8ff0276edb9d8be4f5e66ad2b37d/prow/cluster/crier_deployment.yaml#L15

cjwagner: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/346
ACK about L16

PR #365, opened 2020-05-26
Add required --github-token-path to hook/tide.

michelle192837:
As per flag warning: "missing required flag: please set to --github-token-path=/etc/github/oauth before June 2020"

/assign @fejta @chases2


michelle192837:
Thanks!

/hold cancel

PR #393, opened 2020-06-16
Add knative prow configs

chaodaiG:
- Copy/paste knative prow deployments as well as build cluster deployments configs here, so that auto-bump is handled here
- Add Makefile for manual deployments
- Add README

Postsubmit updating job will be added in knative/test-infra repo later


albertomilan:
There seems to be a typo in the folder structure name: **prow/knaitve**

chaodaiG:
> There seems to be a typo in the folder structure name: **prow/knaitve**

Thanks, fixed and force pushed

PR #398, opened 2020-06-23
exposure-notification jobs use build cluster

chaodaiG:
/hold
Would be nice if we can manually test this job before pushing it

/assign @cjwagner @michelle192837

### DIFF
--- a/prow/prowjobs/google/exposure-notifications-server/en-server.yaml
+++ b/prow/prowjobs/google/exposure-notifications-server/en-server.yaml
@@ -1,6 +1,7 @@
 presubmits:
   google/exposure-notifications-server:
   - name: pull-en-server-release-unit
+    cluster: build-apollo-server
     always_run: true
     decorate: true
     labels:


chaodaiG:
> > Would be nice if we can manually test this job before pushing it
> 
> We can and should test the change. Copy the job into a new presubmit, add a `-canary` suffix to the name, add `skip_report: true`, and then add `cluster: build-apollo-server`. If the canary job runs without issue we can delete it and switch the original job to the new cluster.

Ack, see https://github.com/GoogleCloudPlatform/oss-test-infra/pull/399

sethvargo:
Let me know when it's tested and I'll approve

cjwagner:
> Let me know when it's tested and I'll approve

I don't see any open PRs passing tests so I don't have a green run of the canary presubmit, but here is a failed run that fails in the same way the non-canary presubmit fails: https://oss-prow.knative.dev/view/gcs/oss-prow/pr-logs/pull/google_exposure-notifications-server/659/pull-en-server-release-unit-canary/1275514496298782720

The source code checkout succeeds, the job starts to execute, and results are successfully uploaded, so I'd say its safe to switch the original presubmit the new cluster and delete the canary.

chaodaiG:
There is a passed run now: https://oss-prow.knative.dev/view/gcs/oss-prow/pr-logs/pull/google_exposure-notifications-server/660/pull-en-server-release-unit-canary/1275516887085617152

This PR is good to go @sethvargo 

/unhold

PR #407, opened 2020-06-25
Add knative prow core and plugins config here

chaodaiG:
The original copies still exist in knative/test-infra repo so it won't upset knative prow. Following steps:

1. Update https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/tests/jobs_test.go to ensure no job can use knative prow cluster for any job
1. Add knative prow as a build cluster
1. Stop updating core and plugins from knative prow
1. Update updateconfig plugin config in oss-test-infra prow, let it update knative prow core and plugin config
1. Create a PR with dummy commits, just to verify updateconfig plugin works
1. Cleanup in knative/test-infra repo


cjwagner:
We should have a step between
> 4. Update updateconfig plugin config in oss-test-infra prow, let it update knative prow core and plugin config
> 5. Cleanup in knative/test-infra repo

to PR some dummy changes to the configs (e.g. add a comment) to make sure they are updated properly.

Also I think

> 1. Stop updating core and plugins from knative prow

should happen shortly before 4 to minimize the time during which configs are not automatically updated. We should check if there have been changes to the configs in knative/test-infra and resync if needed.

chaodaiG:
> We should have a step between
> 
> > 1. Update updateconfig plugin config in oss-test-infra prow, let it update knative prow core and plugin config
> > 2. Cleanup in knative/test-infra repo
> 
> to PR some dummy changes to the configs (e.g. add a comment) to make sure they are updated properly.
> 
> Also I think
> 
> > 1. Stop updating core and plugins from knative prow
> 
> should happen shortly before 4 to minimize the time during which configs are not automatically updated. We should check if there have been changes to the configs in knative/test-infra and resync if needed.

Agree on both.

chaodaiG:
> > We should have a step between
> > > 1. Update updateconfig plugin config in oss-test-infra prow, let it update knative prow core and plugin config
> > > 2. Cleanup in knative/test-infra repo
> > 
> > 
> > to PR some dummy changes to the configs (e.g. add a comment) to make sure they are updated properly.
> > Also I think
> > > 1. Stop updating core and plugins from knative prow
> > 
> > 
> > should happen shortly before 4 to minimize the time during which configs are not automatically updated. We should check if there have been changes to the configs in knative/test-infra and resync if needed.
> 
> Agree on both.

Updated description to reflect

PR #409, opened 2020-06-25
Updateconfig plugin updates knative prow configs from oss-test-infra prow

chaodaiG:
This is step 4 as described in https://github.com/GoogleCloudPlatform/oss-test-infra/pull/407

Shouldn't merge until steps 1-3 are done
/hold

### DIFF
--- a/prow/oss/plugins.yaml
+++ b/prow/oss/plugins.yaml
@@ -20,6 +20,16 @@ config_updater:
       name: plugins
     prow/prowjobs/**/*.yaml:
       name: job-config
+    prow/knaitve/config.yaml: # For knative prow
+      name: config
+      clusters: 
+        knative-prow-trusted:
+        - default
+    prow/knative/plugins.yaml: # For knative prow
+      name: plugins
+      clusters: 
+        knative-prow-trusted:
+        - default
 
 approve:
 - repos:
--- a/prow/oss/plugins.yaml
+++ b/prow/oss/plugins.yaml
@@ -20,14 +20,14 @@ config_updater:
       name: plugins
     prow/prowjobs/**/*.yaml:
       name: job-config
-    prow/knaitve/config.yaml: # For knative prow
+    prow/knative/config.yaml: # For knative prow
       name: config
-      clusters: 
+      clusters:
         knative-prow-trusted:
         - default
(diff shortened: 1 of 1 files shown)


prow/oss/plugins.yaml
```diff
@@ -20,6 +20,16 @@ config_updater:
       name: plugins
     prow/prowjobs/**/*.yaml:
       name: job-config
+    prow/knaitve/config.yaml: # For knative prow
+      name: config
+      clusters: 
+        knative-prow-trusted:
```
fejta, line 26: Where do these cluster keys come from?

chaodaiG: It'll be from `kubeconfig` secret, it'll need to be configured by oss-test-infra prow admin

fejta: The kubeconfig for scheduling jobs? This isn't scheduling jobs though.

chaodaiG: Yeah it's not for scheduling jobs, the `kubeconfig` is mentioned as documented in updateconfig plugin:
```
# specify the clusters and namespaces that the configmap targets
      # which requires that either --build-cluster or --kubeconfig arg is enabled for Hook
      # https://github.com/kubernetes/test-infra/blob/master/prow/getting_started_deploy.md#run-test-pods-in-different-clusters
      # if not set or empty, it uses the cluster where prow components are running
      # and the specified namespace(s)
      # if defined, the above namespace will be ignored
      clusters: 
        others:
        - namespace1
```

My understanding is it reuses the same kubeconfig for hook
https://github.com/kubernetes/test-infra/tree/master/prow/plugins/updateconfig

prow/oss/plugins.yaml
```diff
@@ -20,6 +20,16 @@ config_updater:
       name: plugins
     prow/prowjobs/**/*.yaml:
       name: job-config
+    prow/knaitve/config.yaml: # For knative prow
```
cjwagner, line 23: typo

chaodaiG: Fixed, thanks

chaodaiG:
Everything is ready for test
/unhold

I'll monitor the situation

PR #411, opened 2020-06-26
mirror en-server config to en-verification-server

crwilcox:
Seems updates were made to en-server. Trying to keep the two repositories test runner in sync.

### DIFF
--- a/prow/prowjobs/google/exposure-notifications-verification-server/en-verification-server.yaml
+++ b/prow/prowjobs/google/exposure-notifications-verification-server/en-verification-server.yaml
@@ -1,6 +1,7 @@
 presubmits:
   google/exposure-notifications-verification-server:
   - name: pull-en-server-release-unit
+    cluster: build-apollo-server
     always_run: true
     decorate: true
     labels:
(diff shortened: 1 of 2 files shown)


crwilcox:
@sethvargo I think we need the build cluster bits but unsure also. As they were copied I didn't see a need to diverge at this point. Currently prow is failing to run database tests for verification though.

sethvargo:
Oh I don't have perms to approve. @michelle192837

prow/prowjobs/google/exposure-notifications-verification-server/en-verification-server.yaml
```diff
@@ -15,10 +16,21 @@ presubmits:
...
           # We run docker-in-docker which requires privileged mode
           privileged: true
         resources:
           requests:
             cpu: 7
             memory: '16Gi'
+      volumes:
+      - name: gcs-access-service-account
+        secret:
+          secretName: gcs-access-service-account
```
cjwagner, line 36: nit: Consider using GKE workload identity to avoid using a service account key here.

PR #416, opened 2020-06-30
Mount kubeconfig secret for hook

chaodaiG:
Passing `kubeconfig`  as a parameter for hook deployment


cjwagner:
We need to supply the flag as well.

chaodaiG:
> We need to supply the flag as well.

Good catch, added

PR #426, opened 2020-07-08
Update testgrid bazel image to 2.1.0->2.2.0 variant

spiffxp:
Image built from https://github.com/kubernetes/test-infra/pull/18202


spiffxp:
/close
Actually let's try going to 3.0.0 directly, opened https://github.com/GoogleCloudPlatform/oss-test-infra/pull/427

PR #438, opened 2020-07-13
Use ci config for testing

fejta:


fejta:
Will you please review this @chases2 ?

Issue #437, opened 2020-07-13
testgrid config validation failing


BenTheElder:
testgrid config validation(`pull-oss-test-infra-check-testgrid-config`)  is now failing one expecting some fields to be non-zero.

for now to bump prow I overrode the job, but we should fix this.

it's not clear if this is intended and those fields should be changed in the config, or some bad defaulting or ...?

/assign @michelle192837

BenTheElder:
https://oss-prow.knative.dev/view/gcs/oss-prow/logs/post-oss-test-infra-upload-testgrid-config/1282747239776653312

we've also been failing to upload config as a result

nareddyt:
ESPv2 is also blocked by this issue. I took a quick look, but could not find the culprit PR in testgrid. I assume we need to add some annotations to our jobs, but I'm unsure which ones.

michelle192837:
/assign @chases2 

Sean is looking into this (thanks Sean!)

chases2:
After the merge, these tests seem to be passing
(ex. https://github.com/GoogleCloudPlatform/oss-test-infra/pull/438 was failing, but after a retest is passing again)

BenTheElder:
thank you!

PR #440, opened 2020-07-14
b/157921600: Add test case of using app engine as backend

TAOXUY:


TAOXUY:
The broken pull-oss-test-infra-check-testgrid-config has nothing to do with this pr and suspect some base config is incorrect.

nareddyt:
Seems related to #437

PR #458, opened 2020-08-02
Add a regex in spyglass to highlight go build OOM error

chizhg:
When this error happens, the log file is usually very large, so it would be helpful to highlight the error line in Spyglass.

/cc @chaodaiG

### DIFF
--- a/prow/knative/config.yaml
+++ b/prow/knative/config.yaml
@@ -86,21 +86,25 @@ deck:
           - 'timeout waiting for pods to come up'
           # Prow job killed
           - 'Entrypoint received interrupt: terminated'
+          # Go build failure due to OOM
+          - '"go build": signal: killed'
           # Code is outdated
           # TODO(chizhg): this error message should be defined by the common test infrastructure
           - 'Please run .*hack/update-codegen.sh'
           # Metrics are logged only for Pods in a failed test.
           - 'Metrics logs'
           # Compile error - unused import.
-          - "imported and not used:"
+          - 'imported and not used:'
           # Compile error - unused variable.
-          - "declared but not used"
+          - 'declared but not used'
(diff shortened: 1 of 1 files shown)


prow/knative/config.yaml
```diff
@@ -86,6 +86,8 @@ deck:
           - 'timeout waiting for pods to come up'
           # Prow job killed
           - 'Entrypoint received interrupt: terminated'
+          # Go build failure due to OOO
```
chaodaiG, line 89: What's OOO? Do you mean OOM?

chizhg: Sorry, yes... Fixed.

chizhg:
By the way I added a new regex `too many arguments in call to ` to catch another compile error.

prow/knative/config.yaml
```diff
@@ -86,6 +86,8 @@ deck:
           - 'timeout waiting for pods to come up'
           # Prow job killed
           - 'Entrypoint received interrupt: terminated'
+          # Go build failure due to OOM
+          - "\"go build\": signal: killed"
```
chaodaiG, line 90: Would this work?
```suggestion
          - '"go build": signal: killed'
```

chizhg: Yeah it should be working. Updated.

prow/knative/config.yaml
```diff
@@ -99,6 +101,8 @@ deck:
           - "undefined:"
           # Compile error - not enough arguments.
           - "not enough arguments in call to "
+          # Compile error - too many arguments.
+          - "too many arguments in call to "
```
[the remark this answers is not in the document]
chizhg: Good point. Changed them all to use single quotes.

PR #461, opened 2020-08-10
En server: run performance test every 3 hours

chaodaiG:
Run performance test more frequently so that we can get a better sense of moving average changes as in: https://mako.dev/benchmark?benchmark_key=4790091207671808

/cc @sethvargo @MushuEE

### DIFF
--- a/prow/prowjobs/google/exposure-notifications-server/en-server.yaml
+++ b/prow/prowjobs/google/exposure-notifications-server/en-server.yaml
@@ -35,7 +35,7 @@ presubmits:
         secret:
           secretName: gcs-access-service-account
 periodics:
-  - cron: "0 8 * * *"  # Run at 00:00 PST
+  - cron: "* */3 * * *"  # Run every 3 hours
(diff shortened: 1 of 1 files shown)


prow/prowjobs/google/exposure-notifications-server/en-server.yaml
```diff
@@ -35,7 +35,7 @@ presubmits:
         secret:
           secretName: gcs-access-service-account
 periodics:
-  - cron: "0 8 * * *"  # Run at 00:00 PST
+  - cron: "* */3 * * *"  # Run every 3 hours
```
MushuEE, line 38: Should this be `0 */3 * * *`? I don't know cron super well but I am not sure if the * in minutes would mean every minute every 3rd hour

chaodaiG: You are right. It turns out that I'm even less familiar with cron than you are :) 
Fixed

PR #465, opened 2020-08-12
Update kubekins version

TAOXUY:
The old image blocks deploying anthos.


TAOXUY:
> Can you quickly summarize why the new one is needed?

It has the newer gcloud components.

PR #479, opened 2020-08-21
Fix misspelling

spiffxp:
ref: https://github.com/kubernetes/test-infra/pull/18895#issuecomment-678555526


MushuEE:
Does the bot auto rebase?

chases2:
Thanks for finding this bug! I'll look into getting the transfigure script fixed to treat this properly

PR #495, opened 2020-09-02
En verification server: add canary e2e test job

chaodaiG:
/assign @sethvargo

### DIFF
--- a/prow/prowjobs/google/exposure-notifications-verification-server/en-verification-server.yaml
+++ b/prow/prowjobs/google/exposure-notifications-verification-server/en-verification-server.yaml
@@ -36,6 +36,7 @@ presubmits:
   - name: pull-en-verification-server-release-e2e-canary
     cluster: build-apollo-server
     always_run: false
+    optional: true
     decorate: true
     labels:
       preset-dind-enabled: "true"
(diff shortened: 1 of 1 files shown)


prow/prowjobs/google/exposure-notifications-verification-server/en-verification-server.yaml
```diff
@@ -33,3 +33,32 @@ presubmits:
           requests:
             cpu: 7
             memory: '16Gi'
+  - name: pull-en-verification-server-release-e2e-canary
+    cluster: build-apollo-server
+    always_run: true
```
sethvargo, line 38: I thought we werent running this all the time and were going to do a fixed interval instead?

chaodaiG: Sorry forgot to mention, this is for faster turnaround on PRs. Once the PR confirmed working I'll set the periodic runs right away

sethvargo: Will this fail our existing PRs though? Is there a way to mark it as "allow failure"?

chaodaiG: Yeah good catch, changed it to not always run

chaodaiG: Also set optional to true so it's not blocking

PR #506, opened 2020-09-10
En server: email notify if prow jobs failed

chaodaiG:
The alerts will be sent to a google group, so far I'm the only one in the group

/cc @sethvargo @icco


icco (approved):
/lgtm

PR #513, opened 2020-09-16
Move plank and sinker into their own files

fejta:
Also create controller-manager with 0 replicas.
Will soon set plank to 0 and controller-manager to 1 replicas.


fejta:
@chaodaiG @michelle192837 can I get a review here? Thanks!

PR #522, opened 2020-09-29
Knative prow: default job cpu limit to 10

chaodaiG:
Knative serving compilation is CPU intensive, the e2e jobs transiently need to use up to 10 CPUs and get a great performance gain (from 7 minutes to 2 minutes). And the CPU spike only in the second half, which is  ~ 1 minute out of 30 - 50 minutes. So the chance of multiple jobs spiking the exactly the same time is not very high

### DIFF
--- a/prow/knative/cluster/200-cpu-limitrange.yaml
+++ b/prow/knative/cluster/200-cpu-limitrange.yaml
@@ -19,8 +19,6 @@ metadata:
   namespace: test-pods
 spec:
   limits:
-    - default:
-        cpu: 8000m
-      defaultRequest:
+    - defaultRequest:
         cpu: 2000m
       type: Container
--- a/prow/knative/cluster/200-cpu-limitrange.yaml
+++ b/prow/knative/cluster/200-cpu-limitrange.yaml
@@ -20,7 +20,7 @@ metadata:
 spec:
   limits:
     - default:
-        cpu: 4000m
+        cpu: 8000m
       defaultRequest:
         cpu: 2000m
       type: Container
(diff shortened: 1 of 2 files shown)


prow/knative/cluster/200-cpu-limitrange.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
 spec:
   limits:
     - default:
-        cpu: 4000m
+        cpu: 8000m
```
coryrc, line 23: why 8?
"16" would allow all usage

coryrc: Also, as JR points out, the recommendation is to have no CPU limit; if we think deleting that gets us that, we should do that. If not, "16" should be fine?

chaodaiG: When I did the test, setting it at 16 somehow failed to have a pod scheduled. Removing limit is ok too, but I'm not sure how delete works with `kubectl apply`, aka, it may be no op if we simply delete the cpu limit

spiffxp: If you're scheduling to a 16 CPU node, not all of that CPU is allocatable - some of it will be set aside for node components.

We ran into this with kind jobs attempting to schedule "the whole node" on an 8 CPU node build cluster.  Scheduling at 7300m/7400m often resulted in no nodes being available.  Scheduling for 7 worked.

coryrc: Hmm, let's try deleting and see if it works? We can add it back with a different value if absolutely necessary

coryrc: @spiffxp sure you are mixing request and limit?

chaodaiG: I just did a test on my toy cluster, it does get updated when I remove the line of cpu limit, so removing cpu limit now

PR #530, opened 2020-10-05
Update Knative Prow to use Google-managed SSL certificates

chizhg:
Update Knative Prow to use Google-managed SSL certificates, similar as all other Prow instances we manage. This will prevent the SSL certificates to be expired since Google Cloud can renew it automatically.

I have switched Knative Prow to use the new certificate and it's working well without downtime.

/cc @chaodaiG @coryrc 

FYI @peterfeifanchen @albertomilan

### DIFF
--- a/prow/knative/cluster/303-prow-knative-dev_managedcertificate.yaml
+++ b/prow/knative/cluster/303-prow-knative-dev_managedcertificate.yaml
@@ -1,4 +1,4 @@
-# Copyright 2019 The Knative Authors
+# Copyright 2020 The Knative Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


albertomilan:
Is this change self-contained or are we expecting other changes in order to prevent any outage?

chaodaiG:
Awesome, thanks @chizhg !

/lgtm
/approve

chizhg:
> Is this change self-contained or are we expecting other changes in order to prevent any outage?

https://cloud.google.com/load-balancing/docs/ssl-certificates/google-managed-certs#migrating-ssl needs to be followed for the migration so this change is not purely self-contained.

But I have manually done the migration, and the file changes I committed in this PR reflect the final state, so it's safe 😃

albertomilan:
Thanks for the explanation.

/lgtm

PR #533, opened 2020-10-07
set tide to ignore unknown branch status

n3wscott:
Ref: https://github.com/knative/pkg/issues/1785

### DIFF
--- a/prow/knative/config.yaml
+++ b/prow/knative/config.yaml
@@ -192,7 +192,5 @@ tide:
   rebase_label: tide/merge-method-rebase
   merge_label: tide/merge-method-merge
   context_options:
-    # Use branch protection options to define required and optional contexts
-    from-branch-protection: true
     # Treat unknown contexts as optional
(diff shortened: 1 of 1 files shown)


prow/knative/config.yaml
```diff
@@ -191,3 +191,8 @@ tide:
   squash_label: tide/merge-method-squash
   rebase_label: tide/merge-method-rebase
   merge_label: tide/merge-method-merge
+  context_options:
+    # Use branch protection options to define required and optional contexts
+    from-branch-protection: true
```
mattmoor, line 196: We do not want this

mattmoor: Right now many places all we require it tide :)

chaodaiG:
I'm not sure if this solves problem, since tide doesn't understand github action at all, the current behavior of merging blocked by github actions is probably due to github prevents PR from being merged. But it doesn't hurt to give it a shot

/lgtm
/approve

PR #536, opened 2020-10-11
Remove branch-protection config since they are not used for automation here

chizhg:


prow/knative/config.yaml
```diff
@@ -125,21 +125,6 @@ prowjob_namespace: default
 pod_namespace: test-pods
 log_level: debug
 
-branch-protection:
-  orgs:
-    knative:
-      # Protect all branches in knative
-      # This means all prow jobs with "always_run" set are required
```
chaodaiG, line 132: Does removing this means `always_run` will not work any more?

chizhg: This part of config is only consumed by the [branchprotector](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/branchprotector) tool and we have never configured it.
Also since this part of config is independent from the core config, it can be moved to elsewhere as what I did in https://github.com/knative/test-infra/pull/2496

fejta (approved):

PR #540, opened 2020-10-12
Expand regexp for junit files in Spyglass

albertomilan:
Nested folders are a valid use case which other tools are prepared for and Spyglass was omitting internal inspection. This was causing inconsistent behaviors and creating mismatching signals on researches

### DIFF
--- a/prow/knative/config.yaml
+++ b/prow/knative/config.yaml
@@ -110,6 +110,7 @@ deck:
     - lens:
         name: junit
       required_files:
+      # Nested folders is a valid use case  
       - artifacts(/.*/|/)junit.*\.xml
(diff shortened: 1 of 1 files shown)


prow/knative/config.yaml
```diff
@@ -110,7 +110,7 @@ deck:
     - lens:
         name: junit
       required_files:
-      - artifacts/junit.*\.xml
+      - artifacts(/.*/|/)junit.*\.xml
```
chaodaiG, line 113: Can we have a comment here explaining the reason?

albertomilan: I've included a brief comment on top

Issue #474, opened 2020-08-18
Onboard secrets-store-csi-driver-provider-gcp


colinman:
We would like to onboard https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp to oss-test-infra to execute end-to-end tests for the GCP Secret Manager Secrets Store CSI Driver Provider.

As discussed offline, there is some setup work required to configure the Github webhooks.

cjwagner:
Once [this PR](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/475) merges and [`google-oss-robot`](https://github.com/google-oss-robot) has been granted admin access to your repo then it is safe to add an entry to the `managed_webhooks.org_repo_config` section of [`config.yaml`](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/oss/config.yaml). In particular: 
```yaml
managed_webhooks:
    respect_legacy_global_token: true
    org_repo_config:
        # Add this VV
        GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp:
            token_created_after: 2020-08-18T00:10:00Z
```

cjwagner:
Please use `prow/oss/create-build-cluster.sh` once (https://github.com/GoogleCloudPlatform/oss-test-infra/pull/478) merges.

colinman:
Prow tests are running and passing in secrets-store-csi-driver-provider-gcp. Example PR: https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/pull/63

PR #547, opened 2020-10-14
Add GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp Prow job

colinman:
Part of onboarding Secret Manager CSI Driver as part of https://github.com/GoogleCloudPlatform/oss-test-infra/issues/474

### DIFF
--- a/prow/oss/config.yaml
+++ b/prow/oss/config.yaml
@@ -81,7 +81,7 @@ managed_webhooks:
   # Config for orgs and repos that have been onboarded to this Prow instance.
   org_repo_config:
     GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp:
-      token_create_after: 2020-10-13T00:10:00Z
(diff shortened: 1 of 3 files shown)


tam7t (approved):

e-blackwelder (approved):
I noticed you're using `GOOGLE_APPLICATION_CREDENTIALS` -- I think the term for that is something like "service account secrets/keys". While it works, workload identity is the preferred method of authenticating service accounts (for example, since workload identity doesn't use `secret` files, they can't be leaked accidentally). For an overview, see:
https://github.com/kubernetes/test-infra/tree/master/workload-identity#overview

To get workload identity working, you'll first need a (new) kubernetes service account, and then someone (probably the [test-infra oncaller](https://storage.googleapis.com/test-infra-oncall/oncall.html) at the time) will help you bind the cloud and kubernetes accounts -- as per the doc. Feel free to reach out!

For now though, this change looks fine.

PR #570, opened 2020-10-23
Update container registry for KUnit GoB instance

bjh83:
We had to migrate our GCP project containing our container registry for
our Prow Docker image recently, so update the corresponding container
registry location in Prow.


chases2 (approved):
Consider adding an OWNERS file to this directory containing users with knowledge about this ProwJob! It will let you approve changes to this job and gives a potential contact if there's an issue with this job.

chases2:
/hold
(unhold at will)

PR #575, opened 2020-10-24
Add OWNERS file for KUnit GoB instance

bjh83:
It was suggested[1] that we take ownership over the Prow config for KUnit
GoB since most Prow people are not familiar with it. (It also helps the
Prow people find who to talk to if something goes wrong with KUnit Prow
config.)

Link[1]: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/570#pullrequestreview-516016316

### DIFF
--- a/prow/prowjobs/kunit-review.googlesource.com/linux/OWNERS
+++ b/prow/prowjobs/kunit-review.googlesource.com/linux/OWNERS
@@ -1,0 +1,4 @@
+approvers:
+- bjh83
+- dlatypov
+- sulix


bjh83:
> The following users are mentioned in OWNERS file(s) but are untrusted for the following reasons. One way to make the user trusted is to add them as members of the GoogleCloudPlatform org. You can then trigger verification by writing `/verify-owners` in a comment.
> 
> * bjh83
>   
>   * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
> * dlatypov
>   
>   * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.
> * sulix
>   
>   * User is not a member of the org. User is not a collaborator. Satisfy at least one of these conditions to make the user trusted.

Not sure if this was considered when [#570 (review)](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/570#pullrequestreview-516016316) was suggested. However, just to show we are legit, you can see we are all members of the KUnit team inside of Google here: http://team/1720127317258

michelle192837 (approved):

e-blackwelder (approved):

fejta (approved):
/ok-to-test

PR #603, opened 2020-11-16
Configure cherrypicker plugin for Knative Prow

chizhg:
Configure cherrypicker plugin for Knative Prow

Fixes: https://github.com/knative/test-infra/issues/2536

### DIFF
--- a/prow/knative/cluster/500-cherrypicker.yaml
+++ b/prow/knative/cluster/500-cherrypicker.yaml
@@ -1,0 +1,77 @@
+# Copyright 2020 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# cherrypicker plugin
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: default
+  name: cherrypicker
+  labels:
+    app: cherrypicker
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: cherrypicker
+  template:
+    metadata:
+      labels:
+        app: cherrypicker
+    spec:
+      terminationGracePeriodSeconds: 180
+      containers:
+      - name: cherrypicker
+        image: gcr.io/k8s-prow/cherrypicker:v20201109-a36ed74f01
+        args:
+        - --dry-run=false
+        - --use-prow-assignments=false
+        - "--labels=cla: yes"
+#        - --create-issue-on-conflict
+        - --github-endpoint=http://ghproxy
+        - --github-endpoint=https://api.github.com
+        ports:
+        - name: http
+          containerPort: 8888
+        volumeMounts:
+        - name: hmac
+          mountPath: /etc/webhook
+          readOnly: true
+        - name: oauth
+          mountPath: /etc/github
+          readOnly: true
+      volumes:
+      - name: hmac
+        secret:
+          secretName: hmac-token
+      - name: oauth
+        secret:
+          secretName: oauth-token
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/part-of: prow
+    app: cherrypicker
+  namespace: default
+  name: cherrypicker
+spec:
+  selector:
+    app: cherrypicker
+  ports:
+  - name: http
+    port: 80
+    targetPort: http
+  type: ClusterIP


chizhg:
/cc @n3wscott @chaodaiG 
FYI @mattmoor @dprotaso @vaikas

chaodaiG:
This uses the default service account, which is generally not preferable as it's very powerful. Any chance it can use a dedicated service account with certain RBAC permissions?

chizhg:
> This uses the default service account, which is generally not preferable as it's very powerful. Any chance it can use a dedicated service account with certain RBAC permissions?

Hmm I did a quick search, in https://stackoverflow.com/questions/52995962/kubernetes-namespace-default-service-account#:~:text=By%20default%2C%20the%20default%20service,appropriate%20permissions%20to%20do%20that. it said `By default, the default service account in a namespace has no permissions other than those of an unauthenticated user.`, so it should be fine to use the default service account?

chaodaiG:
> > This uses the default service account, which is generally not preferable as it's very powerful. Any chance it can use a dedicated service account with certain RBAC permissions?
> 
> Hmm I did a quick search, in https://stackoverflow.com/questions/52995962/kubernetes-namespace-default-service-account#:~:text=By%20default%2C%20the%20default%20service,appropriate%20permissions%20to%20do%20that. it said `By default, the default service account in a namespace has no permissions other than those of an unauthenticated user.`, so it should be fine to use the default service account?

That's what I assumed as well, but it seemed not the case in prow. For example https://github.com/istio/test-infra/blob/master/prow/cluster/sinker_deployment.yaml, istio prow/sinker uses default service account, but it apparently needs pod deletion permission at least

chizhg:
> > > This uses the default service account, which is generally not preferable as it's very powerful. Any chance it can use a dedicated service account with certain RBAC permissions?
> > 
> > 
> > Hmm I did a quick search, in https://stackoverflow.com/questions/52995962/kubernetes-namespace-default-service-account#:~:text=By%20default%2C%20the%20default%20service,appropriate%20permissions%20to%20do%20that. it said `By default, the default service account in a namespace has no permissions other than those of an unauthenticated user.`, so it should be fine to use the default service account?
> 
> That's what I assumed as well, but it seemed not the case in prow. For example https://github.com/istio/test-infra/blob/master/prow/cluster/sinker_deployment.yaml, istio prow/sinker uses default service account, but it apparently needs pod deletion permission at least

In the sinker case, I think it connects to the Prow and build clusters via `kubeconfig` - https://github.com/istio/test-infra/blob/master/prow/cluster/sinker_deployment.yaml#L22, which grants it super power on the `test-pods` namespace of these clusters.

chaodaiG:
> > > > This uses the default service account, which is generally not preferable as it's very powerful. Any chance it can use a dedicated service account with certain RBAC permissions?
> > > 
> > > 
> > > Hmm I did a quick search, in https://stackoverflow.com/questions/52995962/kubernetes-namespace-default-service-account#:~:text=By%20default%2C%20the%20default%20service,appropriate%20permissions%20to%20do%20that. it said `By default, the default service account in a namespace has no permissions other than those of an unauthenticated user.`, so it should be fine to use the default service account?
> > 
> > 
> > That's what I assumed as well, but it seemed not the case in prow. For example https://github.com/istio/test-infra/blob/master/prow/cluster/sinker_deployment.yaml, istio prow/sinker uses default service account, but it apparently needs pod deletion permission at least
> 
> In the sinker case, I think it connects to the Prow and build clusters via `kubeconfig` - https://github.com/istio/test-infra/blob/master/prow/cluster/sinker_deployment.yaml#L22, which grants it super power on the `test-pods` namespace of these clusters.

Not the prow control plane cluster though, I don't think `kubeconfig` contains prow control plane cluster secret

chizhg:
> > > > > This uses the default service account, which is generally not preferable as it's very powerful. Any chance it can use a dedicated service account with certain RBAC permissions?
> > > > 
> > > > 
> > > > Hmm I did a quick search, in https://stackoverflow.com/questions/52995962/kubernetes-namespace-default-service-account#:~:text=By%20default%2C%20the%20default%20service,appropriate%20permissions%20to%20do%20that. it said `By default, the default service account in a namespace has no permissions other than those of an unauthenticated user.`, so it should be fine to use the default service account?
> > > 
> > > 
> > > That's what I assumed as well, but it seemed not the case in prow. For example https://github.com/istio/test-infra/blob/master/prow/cluster/sinker_deployment.yaml, istio prow/sinker uses default service account, but it apparently needs pod deletion permission at least
> > 
> > 
> > In the sinker case, I think it connects to the Prow and build clusters via `kubeconfig` - https://github.com/istio/test-infra/blob/master/prow/cluster/sinker_deployment.yaml#L22, which grants it super power on the `test-pods` namespace of these clusters.
> 
> Not the prow control plane cluster though, I don't think `kubeconfig` contains prow control plane cluster secret

I'm not sure how Istio Prow cluster was setup, but none of its deployments are using fine-grained service account:
- https://github.com/istio/test-infra/blob/master/prow/cluster/hook_deployment.yaml#L21
- https://github.com/istio/test-infra/blob/master/prow/cluster/horologium_deployment.yaml

Maybe worth to dig it through how it works, but since Knative Prow is using the dedicated service account for each deployment - https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/knative/cluster/400-sinker.yaml#L29, I don't think it's a big problem to use the default service account here.

chaodaiG:
sgtm
/lgtm

chizhg:
Hmm.. `tide Pending — Not mergeable.`

PR #618, opened 2020-11-17
Fix yamllint job scanning path

chaodaiG:
It was a bad copy/paste from somewhere else, ymal lint check should be on the entire repo


chaodaiG:
This time tested with pj-on-kind, should be good

/cc @michelle192837 @chases2

michelle192837 (approved):

PR #619, opened 2020-11-17
Enable retitle plugin in knative repositories

albertomilan:


chaodaiG:
This is just for knative org, do you also want it in knative-sandbox and other org/repo?

/lgtm
/approve
/hold
Feel free unhold if you think knative is sufficient

albertomilan:
We have not received any request to enable it outside of knative. I'd rather be defensive and limit plugins to customer requests.
If plugins should be enabled by default, I would argue why do we have to configure them? (I would understand a blocklist for particular limitations, but not manual enablements)

PR #631, opened 2020-12-01
Add testgrid alerts for important prow maintenance jobs

chaodaiG:


prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -122,6 +124,8 @@ postsubmits:
     annotations:
       testgrid-dashboards: googleoss-test-infra
       testgrid-tab-name: knative-prow-deploy
+      testgrid-alert-email: k8s-infra-oncall@google.com
+      testgrid-num-failures-to-alert: '1'
```
fejta, line 128: 1 will not distinguish an outage from flakiness. I'd vote for making these 2 or 3.

fejta (approved):
/hold

chases2 (approved):

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
chaodaiG: Makes sense, up postsubmit to be 2 and periodic to be 3

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -169,6 +175,8 @@ postsubmits:
     annotations:
       testgrid-dashboards: googleoss-test-infra
       testgrid-tab-name: hmacs-reconciliation
+      testgrid-alert-email: k8s-infra-oncall@google.com
+      testgrid-num-failures-to-alert: '2'
```
cjwagner, line 179: nit: We might want to know when the hmac reconciler fails since that would probably need to be rerun even for a single transient failure. Otherwise there would be a missing webhook or a secret that was not rotated and it wouldn't be addressed until the next change to the webhook config which could be a long while.

cjwagner (approved):
/hold

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
chaodaiG: Done

cjwagner (approved):

PR #632, opened 2020-12-01
Prow alert emails send to oncall

chaodaiG:


prow/knative/cluster/monitoring/secrets/alertmanager-prow_secret.yaml
```diff
@@ -32,11 +32,16 @@ stringData:
     receivers:
     - name: 'email-alerts'
       email_configs:
-      - to: 'chaodai+prow-alert@google.com'
+      - to: 'k8s-infra-oncall+prow-alert@google.com'
```
cjwagner, line 35: k8s-infra-oncall is not appropriate for these instances. We should use our oncall group or osp-engprod-tools instead.

prow/knative/cluster/monitoring/secrets/alertmanager-prow_secret.yaml
```diff
@@ -21,7 +21,7 @@ stringData:
       group_wait: 30s
       group_interval: 10m
       repeat_interval: 4h
-      receiver: 'no-op-alerts'
+      receiver: 'email-warnings'
```
cjwagner, line 24: This is going to be really noisy since we have a number of alerts that are low priority or inactionable. See the #prow-alerts channel on the k8s slack to see what I mean. Please don't start routing these alerts to us until we've reduced the noise.

cjwagner (changes requested):

prow/knative/cluster/monitoring/secrets/alertmanager-prow_secret.yaml
chaodaiG: Done

prow/knative/cluster/monitoring/secrets/alertmanager-prow_secret.yaml
chaodaiG: sg, removed

e-blackwelder:
Looks reasonable to me (didn't know you could format MDB emails like that). Will let Cole take a look.

cjwagner (approved):
I don't think you need to have both the oncall group and your email given that you are in the oncall group, but it shouldn't really matter. LGTM.

cjwagner:
These secrets will need to be manually edited/deployed since they contains credentials that are not checked in.

chaodaiG:
> These secrets will need to be manually edited/deployed since they contains credentials that are not checked in.

Applied

PR #635, opened 2020-12-03
Remove resultstore service account and jobs.

fejta:
We are no longer using this service


fejta:
@BenTheElder @michelle192837 can I get a review?

BenTheElder:
I see my migration to github notifications over email / gubernator did not go so well in some cases :(

fejta:
> my migration to github notifications 

You are a brave man!

PR #652, opened 2021-01-06
Add missing ssh key secret to autobump testgrid jobs

mpherman2:
/hold
Attempting to debug testgrid autobump. Not sure if this is all that is needed

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
@@ -277,7 +277,7 @@ postsubmits:
       testgrid-num-failures-to-alert: '3'
     spec:
       containers:
-      - image: gcr.io/k8s-testimages/generic_autobump:v20210106-34a4396
+      - image: gcr.io/k8s-testimages/generic_autobump:v20210107-f3f8a3b
         command:
         - /generic_autobump
         args:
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -286,10 +286,16 @@ postsubmits:
...
           readOnly: true
+        - name: ssh
+          mountPath: /root/.ssh
       volumes:
       - name: github
         secret:
           secretName: oauth-token
+      - name: ssh
+        secret:
+          secretName: k8s-ci-robot-ssh-keys
```
cjwagner, line 297: This secret doesn't (and shouldn't) exist on this Prow instance. If an ssh key is needed then we need to use one associated with the bot account for this Prow instance, not k8s-prow.

cjwagner (changes requested):

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
mpherman2: Fixed this so these should be google-oss-robot keys now. I also made changes to the testgrid config to also point to google-oss-robot instead of k8s here: https://github.com/GoogleCloudPlatform/testgrid/pull/324/files, and updated the image to the latest one that has git in it to fix the git error.

cjwagner (approved):
LGTM. `/hold cancel` once the secret exists in the cluster.
/hold

PR #653, opened 2021-01-11
Register 'build-kubeflow' cluster with OSS-Prow.

cjwagner:
Trying out using the `KUBECONFIG` env var rather than merging all kubeconfigs into a single secret.

See the discussions [here](https://github.com/kubernetes/test-infra/pull/18695#discussion_r466087229) and [this issue](https://github.com/kubernetes/test-infra/pull/20417) for context. Our current pattern works reasonably well, but has the downside of not showing which clusters were added in the git history. Switching to using separate secrets and the `KUBECONFIG` envvar will give us the git history we want and avoid secret size limits down the road, but at the cost of making our deployment files harder to read and making the consumers of the kubeconfig implicit rather than explicit. This tradeoff didn't seem worth it at first, but does now in retrospect.

I've already created the new secret in the cluster and confirmed that the key is `kubeconfig`:
```
secret/kubeconfig-build-kubeflow created
```

/assign @chaodaiG

### DIFF
--- a/prow/oss/cluster/crier.yaml
+++ b/prow/oss/cluster/crier.yaml
@@ -32,8 +32,11 @@ spec:
         - --github-token-path=/etc/github/oauth
         - --github-workers=1
         - --job-config-path=/etc/job-config
-        - --kubeconfig=/etc/kubeconfig/config-20201002
         - --kubernetes-blob-storage-workers=1
+        env:
+        # Use KUBECONFIG envvar rather than --kubeconfig flag in order to provide multiple configs to merge.
+        - name: KUBECONFIG
+          value: "/etc/kubeconfig/config-20201002:/etc/build-kubeflow/kubeconfig"
         volumeMounts:
         - name: cookies
           mountPath: /etc/cookies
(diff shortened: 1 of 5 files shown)


prow/oss/cluster/crier.yaml
```diff
@@ -47,6 +50,9 @@ spec:
         - name: kubeconfig
           mountPath: /etc/kubeconfig
           readOnly: true
+        - name: build-kubeflow
```
chaodaiG, line 53: Would it be easier to use `secret.items` like https://github.com/openshift/release/blob/4b6b42276b95d8688c9327158b28aaa5dcdfd2c5/clusters/app.ci/prow/03_deployment/hook.yaml#L231?

cjwagner: I don't see how that would help. All those keys are from the same secret which is not the case for us (and shouldn't be if we want to avoid secret size limits down the road). AFAICT that is just being used to rename the keys.

chaodaiG: Gotcha, overlooked the last line of the openshift config.

PR #659, opened 2021-01-14
Enable tide reviewApprovedRequired on knative-sandbox/net-contour

n3wscott:


prow/knative/config.yaml
```diff
@@ -173,6 +174,18 @@ tide:
     missingLabels: *knative_tide_missing_labels
     author: knative-prow-robot
     reviewApprovedRequired: false
+  - repos:
+    - knative-sandbox/net-contour
+    labels:
+    - lgtm
+    - approved
```
n3wscott, line 181: will github reviews make the approved label? I don't think so, but then will prow add approved label from reviews? I thought that is what @grantr added.

chaodaiG: I think prow will add approved labels, but your change seems to make this repo requires `lgtm` and `approve` label, as well as explicit github code review approval, is this desired?

chaodaiG:
/approve

/hold
If you want to address my nits.

prow/knative/config.yaml
n3wscott: yes, that is the goal

prow/knative/config.yaml
```diff
@@ -173,6 +174,18 @@ tide:
...
+    - "cla: yes"
+    missingLabels: &knative_tide_missing_labels
```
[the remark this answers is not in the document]
n3wscott: oh yeah don't want do to that, good catch.

PR #661, opened 2021-01-15
Expose prow monitoring dashboard

chaodaiG:
The DNS entry was registered to be associated with the global static ip named , this makes grafana exposed on the domain oss-prow-monitoring.knative.dev

### DIFF
--- a/prow/oss/cluster/monitoring/grafana_expose.yaml
+++ b/prow/oss/cluster/monitoring/grafana_expose.yaml
@@ -18,33 +18,32 @@ spec:
     targetPort: nginx
   selector:
     app: grafana
-# TODO(chaodaiG): uncomment this once figure out how to support ingress
-# ---
-# apiVersion: networking.gke.io/v1beta1
-# kind: ManagedCertificate
-# metadata:
-#   name: oss-prow-knative-dev
-#   namespace: prow-monitoring
-# spec:
-#   domains:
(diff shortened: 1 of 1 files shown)


prow/oss/cluster/monitoring/grafana_expose.yaml
```diff
@@ -18,33 +18,32 @@ spec:
...
+    kubernetes.io/ingress.global-static-ip-name: oss-prow-monitoring-ip
+    kubernetes.io/ingress.class: "gce"
+    kubernetes.io/tls-acme: "true"
+    networking.gke.io/managed-certificates: oss-prow-knative-dev
+spec:
+  rules:
+  - host: oss-prow-monitoring.knative.dev
+    http:
+      paths:
+      - path: /monitoring
```
cjwagner, line 46: Please omit this path.

cjwagner (commented):
Please also change the type of the grafana k8s service. It is currently NodePort, but cluster IP should be sufficient.

prow/oss/cluster/monitoring/grafana_expose.yaml
chaodaiG: Good catch, done

chaodaiG:
Switched to ClusterIp as well

cjwagner (approved):

PR #663, opened 2021-01-15
secrets-store-csi-driver-provider-gcp: remove secret

tam7t:
Switching to using workload identity instead.

https://github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/pull/81 binds the `test-jobs/default` k8s SA to the correct GCP SA

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/secrets-store-csi-driver-provider-gcp-config.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/secrets-store-csi-driver-provider-gcp-config.yaml
@@ -5,6 +5,7 @@ presubmits:
     decorate: true           # Decorate with Prow utilities
     always_run: true         # Run for every PR, or only when requested.
     spec:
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/secrets-store-csi-driver-provider-gcp-config.yaml
```diff
@@ -9,13 +9,3 @@ presubmits:
       - image: gcr.io/google.com/cloudsdktool/cloud-sdk:latest
```
fejta, line 9: So long as you make the default k8s SA in the above cluster context use workload identity this works great.

Otherwise you'll need a `serviceAccountName` on this podspec

fejta (approved):
/hold
(run `/hold cancel` if you want this to merge as is)

fejta (approved):
/hold

Note that merging this will cause pods for this job to stay in a pending state until `prow-test-sa` exists. Unhold at that time

colinman (approved):

PR #674, opened 2021-01-21
test(kubeflow): add KFP presubmit tests as optional

Bobgy:
/assign @chaodaiG 

Tests copied from https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubeflow/kubeflow-presubmits.yaml.
Only added kubeflow-pipeline tests, others will be moved later.

### DIFF
--- a/prow/prowjobs/kubeflow/kubeflow-presubmits.yaml
+++ b/prow/prowjobs/kubeflow/kubeflow-presubmits.yaml
@@ -14,3 +14,237 @@ presubmits:
         args:
         - -c
         - cd ./frontend && npm run test:ci:prow
+  - name: kubeflow-pipeline-backend-v2-test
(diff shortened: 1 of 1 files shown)


prow/prowjobs/kubeflow/kubeflow-presubmits.yaml
```diff
@@ -14,3 +14,237 @@ presubmits:
...
+    spec:
+      containers:
+      - image: python:3.6
+        command:
+        - ./test/presubmit-backend-visualization.sh
+
+  # this test is not passing
+  - name: kubeflow-pipeline-multiuser-test
+    cluster: build-kubeflow
+    always_run: false
```
chaodaiG, line 238: optional: true

PR #678, opened 2021-01-25
OSS prow monitoring: update based on all recent improments from k8s prow monitoring stack

chaodaiG:
Includes:
- Major defaulting improvements.
- Improved ghproxy dashboards.
- Stale image alert bug fix.


fejta (approved):
/hold

PR #687, opened 2021-01-26
Configure oss prow grafana exposure to match k8s prow settings

chaodaiG:
These annotations seem to be required for the ingress to behave normally. Also by the way:

- Rename managedcertificate so that it's not the same name as the one used for deck.
- Remove subpath `/monitoring` as it's using a subdomain now.


fejta (approved):
/hold

chaodaiG:
This had been tested and proved to be working, so

/unhold

PR #688, opened 2021-01-29
Only push images when non-cluster files change

fejta:
/assign @chases2 

https://play.golang.org/p/UBRAEDvGSq1


chases2 (approved):

fejta:
Thanks for that terrible regex Sean :)

PR #689, opened 2021-02-01
[ESPv2] Remove GKE version

nareddyt:
I don't think this arg is needed. Our tests keep breaking every month when an old version goes out of date. Let's just remove it.

```
ERROR: (gcloud.container.clusters.create) ResponseError: code=400, message=No valid versions with the prefix "1.17.13" found.
```


nareddyt:
@TAOXUY PTAL

TAOXUY (approved):

PR #692, opened 2021-02-02
Fix prow monitoring stack

chaodaiG:
Fix a couple of errors introduced in https://github.com/GoogleCloudPlatform/oss-test-infra/pull/678

### DIFF
--- a/prow/oss/cluster/monitoring/mixins/prometheus/prometheus.libsonnet
+++ b/prow/oss/cluster/monitoring/mixins/prometheus/prometheus.libsonnet
@@ -10,6 +10,4 @@
 (import 'stale_alerts.libsonnet') +
 (import 'tide_alerts.libsonnet') +
 (import 'prober_alerts.libsonnet') +
-(import 'boskos_alerts.libsonnet') +
-(import 'plank_alerts.libsonnet') +
 (import 'slo_recordrules.libsonnet')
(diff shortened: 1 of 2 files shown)


prow/oss/cluster/monitoring/mixins/prometheus/prometheus.libsonnet
```diff
@@ -10,6 +10,4 @@
 (import 'stale_alerts.libsonnet') +
 (import 'tide_alerts.libsonnet') +
 (import 'prober_alerts.libsonnet') +
-(import 'boskos_alerts.libsonnet') +
-(import 'plank_alerts.libsonnet') +
```
fejta, line 14: We don't want plank alerts?

fejta (approved):
/hold

prow/oss/cluster/monitoring/mixins/prometheus/prometheus.libsonnet
chaodaiG: The plank alert in k8s prow is a heartbeat job, which doesn't exist in oss prow yet. Will add the plank alert once the heartbeat job is up.

k8s plank alert: https://github.com/kubernetes/test-infra/blob/master/config/prow/cluster/monitoring/mixins/prometheus/plank_alerts.libsonnet#L17

PR #693, opened 2021-02-03
Knative prow monitoring: update based on all recent improments from k8s prow monitoring stack

chaodaiG:


e-blackwelder:
/lgtm

... Looks Reasonable To Me.

PR #701, opened 2021-02-16
Replace old autobump job with new, configurable autobump

mpherman2:


prow/knative/knative-autobump-config.yaml
```diff
@@ -0,0 +1,22 @@
...
+upstreamURLBase: "https://raw.githubusercontent.com/kubernetes/test-infra/master"
+includedConfigPaths:
+  - "prow/knative/cluster"
+targetVersion: "upstream"
+extraFiles:
+  - "prow/knative/config.yaml"
+prefixes:
+  - name: "knative-prow"
+    prefix: "gcr.io/k8s-prow/"
+    refConfigFile: "config/prow/cluster/deck_deployment.yaml"
```
chaodaiG, line 19: Can this be defaulted? This tends to break with renames etc.

prow/knative/knative-autobump-config.yaml
```diff
@@ -0,0 +1,22 @@
+---
+gitHubLogin: "google-oss-robot"
+gitHubToken: "/etc/github-token/oauth"
+onCallAddress: "https://storage.googleapis.com/kubernetes-jenkins/oncall.json"
+skipPullRequest: false
+gitHubOrg: "GoogleCloudPlatform"
+gitHubRepo: "oss-test-infra"
+remoteName: "oss-test-infra"
+headBranchName: "autobump-knative-prow"
+upstreamURLBase: "https://raw.githubusercontent.com/kubernetes/test-infra/master"
```
chaodaiG, line 10: Same as below, can this be defaulted?

chaodaiG (commented):
Generally lgtm, I was expecting more defaults, so that it's less prone to future errors

prow/knative/knative-autobump-config.yaml
mpherman2: When we've discussed defaulting in the past, we decided that fields that will always vary for different use cases should not be defaulted. Accidentally forgetting this field when bumping a prefix that is not k8s-prow would cause a very confusing error if we were using defaults.

prow/knative/knative-autobump-config.yaml
mpherman2: Same as below

chaodaiG:
I see, thanks, LGTM, please fix merge conflict then I'll lgtm

PR #714, opened 2021-02-22
Move the knative-sandbox org to CODEOWNERS

mattmoor:
/hold
cc @rhuss @lionelvillard

/assign @chizhg @chaodaiG

### DIFF
--- a/prow/knative/plugins.yaml
+++ b/prow/knative/plugins.yaml
@@ -16,6 +16,7 @@
 approve:
 - repos:
   - "knative"
+  # TODO(mattmoor): Should we disable this for sandbox?
   - "knative-sandbox"
   - "google/knative-gcp"
   - "knative-prow-robot/test-infra"
@@ -25,23 +26,8 @@ approve:
 # Settings for the "lgtm" plugin.
 lgtm:
 - repos:
-  # For now only start with knative-sandbox/net-contour
-  - knative-sandbox/.github
-  - knative-sandbox/downstream-test-go
-  - knative-sandbox/integration
-  - knative-sandbox/knobots
-  - knative-sandbox/kperf
-  - knative-sandbox/net-certmanager
-  - knative-sandbox/net-contour
-  - knative-sandbox/net-http01
-  - knative-sandbox/net-ingressv2
-  - knative-sandbox/net-istio
-  - knative-sandbox/net-kourier
-  - knative-sandbox/sample-controller
-  - knative-sandbox/sample-source
-  - knative-sandbox/reconciler-test
+   - "knative-sandbox"
 #  - "knative"
-#  - "knative-sandbox"
 #  - "google/knative-gcp"
   review_acts_as_lgtm: true
 
--- a/prow/knative/config.yaml
+++ b/prow/knative/config.yaml
@@ -136,27 +136,12 @@ tide:
   queries:
   - orgs:
     - "knative"
-    - "knative-sandbox"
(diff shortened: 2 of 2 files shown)


prow/knative/plugins.yaml
```diff
@@ -16,6 +16,7 @@
 approve:
 - repos:
   - "knative"
+  # TODO(mattmoor): Should we disable this for sandbox?
```
mattmoor, line 19: @chaodaiG @chizhg Please advise on whether I should drop this?

prow/knative/plugins.yaml
```diff
@@ -25,23 +26,8 @@ approve:
...
-  - knative-sandbox/net-certmanager
-  - knative-sandbox/net-contour
-  - knative-sandbox/net-http01
-  - knative-sandbox/net-ingressv2
-  - knative-sandbox/net-istio
-  - knative-sandbox/net-kourier
-  - knative-sandbox/sample-controller
-  - knative-sandbox/sample-source
-  - knative-sandbox/reconciler-test
+   - "knative-sandbox"
```
mattmoor, line 29: I am assuming `repos` is correct mostly based on the commented lines below.  @chizhg @chaodaiG any sanity checking would be wonderful 🙏

chaodaiG: I believe it's correct

mattmoor: There are a few other places that seem to use `repos:` this way, so I figured it might be, but other places have `orgs:` too 😅

prow/knative/plugins.yaml
chaodaiG: Not sure about the motivation. It reacts to `/lgtm`(When lgtm as approve is enabled) and `/approve` comments and approves PR. Keeping this on shouldn't hurt imo

chaodaiG (approved):
Based on my knowledge, this is good to go

prow/knative/plugins.yaml
mattmoor: Well, without OWNERS files it leaves a confusing "NOT APPROVED" message file before the PR is merged anyhow. 😅

mattmoor: See: https://github.com/knative-sandbox/net-contour/pull/434#issuecomment-782245261

prow/knative/plugins.yaml
chaodaiG: You are definitely right about this. Generally prow configs are pretty consistent among each other, but when it comes to each individual plugins, there is no such good cohesion, as each plugin gets to decide the structure for itself, as well as how to interpret them (https://github.com/kubernetes/test-infra/blob/master/prow/plugins/config.go). This imo had become a tech debt and cleaning up is not an easy task.

prow/knative/plugins.yaml
chaodaiG: ah. I see 😅

chaodaiG: If there are still repos don't have OWNERS file, maybe should drop it for now?

mattmoor: The goal would be to drop OWNERS from all repos, right now the CODEOWNERS files coexist, but with this PR the source of truth will cut over to CODEOWNERS as the source of truth and OWNERS will become meaningless.

mattmoor:
/unhold

We can drop the approve plugin in a follow up

PR #732, opened 2021-03-09
test(kubeflow): better run_if_changed conditions

Bobgy:
/assign @chensun @zijianjoy @capri-xiyue 

context: I was working on https://github.com/kubeflow/pipelines/pull/5266 and noticed that some tests should be triggered but they were not, so I added those rules to run_if_changed.


chensun:
/lgtm

Thanks!

PR #742, opened 2021-03-16
[http-pattern-matcher] Add Prow setup for google/http_pattern_matcher

nareddyt:
Adds configurations to use Prow with pre-existing `espv2` test cluster for the repository https://github.com/google/http_pattern_matcher.

Same implementation details as #728.


nareddyt:
cc @qiwzhang @TAOXUY

nareddyt:
cc @shuoyang2016

some background: this is the repository that runs our OSS CI. 
Has configurations for 3 repos we maintain: espv2, grpc-httpjson-transcoder, and this new http_pattern_matcher

qiwzhang (approved):

prow/prowjobs/google/http_pattern_matcher/http-pattern-matcher.yaml
```diff
@@ -0,0 +1,34 @@
+presubmits:
```
chaodaiG, line 1: I can see that the `trigger` plugin is enabled on all `google` repos. For presubmit jobs to be triggered, webhooks need to be setup to be sent to prow. Since the trigger plugin is enabled for entire `google` org and prow isn't working on all repos, I think the webhooks are set up separatedly in each repo under `google`.  If it's not been done yet, you can do it by adding an entry under https://github.com/GoogleCloudPlatform/oss-test-infra/blob/9babc06a371f668163344320eed7178602183618/prow/oss/config.yaml#L97. Please also make sure that google-oss-robot is an admin of google/http_pattern_matcher

nareddyt: I added the webhook config to google/http_pattern_matcher and sent an invitation to `google-oss-robot`.

I don't understand why the `config.yaml` entry needs to be changed. I followed the same approach as #728 (which seems to trigger correctly). Could you clarify if that is needed?

chaodaiG: Thank you!

In the PR description of #728, it mentioned:
> Also configured the webhook, using a similar IP address and secret as our pre-existing project (esp-v2).

The configuration change I mentioned above serves exactly the same purpose as the statement, which is a neat feature of prow that automatically sets up rotating webhooks, so that user doesn't need to go to Github UI to configure webhook.

chaodaiG: The invitation was accepted

nareddyt: Ah got it, neat feature! Done.

chaodaiG (approved):

PR #751, opened 2021-03-19
Knative prow monitoring secret: updated with a script instead of make…

chaodaiG:
… rule

It's pretty hard to have make rules to exit based on wrapped subshell, turn it into a script and it should exit correctly

### DIFF
--- a/prow/knative/cluster/monitoring/Makefile
+++ b/prow/knative/cluster/monitoring/Makefile
@@ -24,12 +24,6 @@ apply-components: get-cluster-credentials
 
 apply: apply-components
 	$(MAKE) -C ./mixins apply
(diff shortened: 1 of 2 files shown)


chaodaiG:
/hold
The secret doesn't exist yet

cjwagner (approved):
The script is more readable too :+1:

cjwagner:
> The secret doesn't exist yet

Should we remove the existing logic in a separate PR in the meantime or will the secret be exist soon?

chaodaiG:
> > The secret doesn't exist yet
> 
> Should we remove the existing logic in a separate PR in the meantime or will the secret be exist soon?

They exist now :)
/unhold

PR #770, opened 2021-04-09
Allow PRs created by knative-prow-updater-robot to be automatically merged

chizhg:
/cc @chaodaiG

### DIFF
--- a/prow/knative/config.yaml
+++ b/prow/knative/config.yaml
@@ -168,9 +168,9 @@ tide:
     - "knative/test-infra"
     labels:
     - "cla: yes"
-    - auto-merge
+    - "skip-review"
     missingLabels: *knative_tide_missing_labels
-    author: knative-prow-robot
+    author: knative-prow-updater-robot
     reviewApprovedRequired: false
   - orgs:
     - "knative-sandbox"


prow/knative/config.yaml
```diff
@@ -172,6 +172,14 @@ tide:
     missingLabels: *knative_tide_missing_labels
     author: knative-prow-robot
     reviewApprovedRequired: false
+  - repos:
+    # Allow PRs for test-infra to be automatically merged without manual approval
+    - "knative/test-infra"
+    labels:
+      - "cla: yes"
```
chaodaiG, line 179: How about adding a label like https://github.com/kubernetes/test-infra/blob/7d5e552c49619d424c4cb20429c8ec0ca1bba750/config/prow/config.yaml#L551, to make this better controllable?

chaodaiG: Because that will allow you to do something like https://github.com/knative/test-infra/pull/2729#issuecomment-816998924

chaodaiG:
Also
/cc @n3wscott

chaodaiG:
ok CC @n3wscott

prow/knative/config.yaml
chizhg: Good point. Updated in https://github.com/knative/test-infra/pull/2746

chaodaiG (approved):

PR #783, opened 2021-04-20
Add private deck deployment to oss prow

chaodaiG:
Depends on #781 , plus not deploy late in the day
/hold

### DIFF
--- a/prow/oss/cluster/deck_private_deployment.yaml
+++ b/prow/oss/cluster/deck_private_deployment.yaml
@@ -1,0 +1,208 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: deck-private
+  labels:
+    app: deck-private
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: deck-private
+  template:
+    metadata:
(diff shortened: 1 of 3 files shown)


chaodaiG:
/unhold
Let's try it out

prow/oss/config.yaml
```diff
@@ -64,10 +64,11 @@ tide:
 plank:
   job_url_prefix_config:
     '*': https://oss-prow.knative.dev/view/
+    'fejtaverse': https://oss-prow-private.knative.dev/view/
```
cjwagner, line 67: nit: This isn't necessary since we aren't setting up any fetaverse jobs.

prow/oss/cluster/deck_private_deployment.yaml
```diff
@@ -0,0 +1,208 @@
+apiVersion: apps/v1
```
cjwagner, line 1: nit: This file contains all the deck private resources, not just the deployment. Doesn't really matter though since we aren't using bazel to deploy.

cjwagner (approved):

PR #789, opened 2021-04-21
Fix a typo in private oss prow deck managed cert config

chaodaiG:


cjwagner (approved):
Oops I definitely saw this while we were pair programming, but when you scrolled back it looked right. I must have been looking at the resource below. :man_facepalming:

PR #790, opened 2021-04-21
Enable inrepoconfig for chaotoppicks org

chaodaiG:


prow/prowjobs/chaotoppicks/great-tools/great-tools.yaml
```diff
@@ -0,0 +1,18 @@
+presubmits:
+  chaotoppicks/great-tools:
+  - name: pull-chaotoppicks-greate-tools-validate-prow-yaml
```
cjwagner, line 3: The jobs for the public repo would need to run in a different cluster from the private build cluster `build`.

prow/oss/config.yaml
```diff
@@ -150,6 +150,13 @@ deck:
       github_orgs:
       - GoogleCloudPlatform
 
+in_repo_config:
+  enabled:
+    chaotoppicks: true
+
+  allowed_clusters:
+    "chaotoppicks": ["build"]
```
cjwagner, line 158: If we're actually going to create a build cluster to test this, please name it something more specific than `build`.

cjwagner (changes requested):
FWIW I'm starting to question the value in setting this all up as a demo.

prow/prowjobs/chaotoppicks/great-private-tools/great-private-tools.yaml
```diff
@@ -0,0 +1,18 @@
+presubmits:
+  chaotoppicks/great-private-tools:
+  - name: pull-chaotoppicks-greate-private-tools-validate-prow-yaml
```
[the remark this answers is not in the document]
chaodaiG: fixed

prow/prowjobs/chaotoppicks/great-private-tools/great-private-tools.yaml
```diff
@@ -0,0 +1,18 @@
+presubmits:
+  chaotoppicks/great-private-tools:
```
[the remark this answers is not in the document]
chaodaiG: added

prow/oss/config.yaml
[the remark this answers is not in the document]
chaodaiG: Added a test

prow/prowjobs/chaotoppicks/great-tools/great-tools.yaml
chaodaiG: Not quite sure whether this needs to be a requirement, but it's a good separation, so done

prow/oss/config.yaml
chaodaiG: done

prow/tests/config_test.go
```diff
@@ -0,0 +1,64 @@
...
+				or = strings.Replace(or, gerritPrefix, "", 1)
+			}
+		}
+		parts := strings.SplitN(or, "/", 2)
+		if len(parts) > 1 {
+			repo += parts[1]
+		}
+		org += parts[0]
+		if !allowedOrgRepo.Has(org) && !allowedOrgRepo.Has(org+"/"+repo) {
+			t.Fatalf("Org/repo not allowed to use private cluster. Want: %q enable inrepoconfig, got: %v", or, allowedOrgRepo.List())
```
cjwagner, line 61: If I understand correctly this validates that in repo config is enabled for any org or repo that allows a private build cluster for in repo config jobs. I don't think this is something we need to check. It doesn't make sense for an org or repo to configure allowed build clusters for inrepo config if they don't even have inrepo config enabled so that should be checked by `checkconfig` if it isn't already.

I think what we actually need to be validating is that no jobs in this repo run in build clusters that include `private` with the exception of the `checkconfig` jobs.

cjwagner (approved):
/hold

prow/tests/config_test.go
cjwagner: If we want to keep this validation it should be fixed to no longer be specific to private repos and we should upstream it.

prow/tests/jobs_test.go
```diff
@@ -83,6 +85,38 @@ func TestTrustedJobs(t *testing.T) {
...
 }
 
+func TestPrivateJobs(t *testing.T) {
+	const (
+		private           = "private"
+		checkconfigPrefix = "gcr.io/k8s-prow/checkconfig:"
+	)
+
+	errorIfNotPermitted := func(t *testing.T, name, cluster string, spec *v1.PodSpec) {
+		if strings.Contains(cluster, private) && (len(spec.Containers) != 1 || !strings.HasPrefix(spec.Containers[0].Image, "gcr.io/k8s-prow/checkconfig:")) {
```
cjwagner, line 95: I don't think this is sufficiently secure. This still allows anyone to create prowjobs for private clusters, they just have to use the checkconfig image. For example someone could create a checkconfig job and make the entrypoint an arbitrary `bash -c` command.
I think it easier and safer to only allow jobs for private clusters to be configured in a specific file/directory that we lock down with OWNERS files.

prow/tests/config_test.go
chaodaiG: I'll just drop this

prow/tests/jobs_test.go
chaodaiG: I see, makes sense. I'll create the dir

chaodaiG: done

cjwagner (approved):

PR #792, opened 2021-04-21
Deploy private deck in default namespace, btw change them to clusterI…

chaodaiG:
…P instead of nodeIP

### DIFF
--- a/prow/oss/cluster/cluster.yaml
+++ b/prow/oss/cluster/cluster.yaml
@@ -70,15 +70,6 @@ spec:
 apiVersion: networking.gke.io/v1beta1
 kind: ManagedCertificate
 metadata:
-  name: oss-prow-private-knative-dev
-  namespace: default
-spec:
-  domains:
-  - oss-prow-private.knative.dev
----
-apiVersion: networking.gke.io/v1beta1
-kind: ManagedCertificate
-metadata:
   name: oss-prow-knative-dev
   namespace: default
 spec:
@@ -94,7 +85,7 @@ metadata:
     kubernetes.io/ingress.global-static-ip-name: "oss-prow"
     kubernetes.io/ingress.class: "gce"
     kubernetes.io/tls-acme: "true"
-    networking.gke.io/managed-certificates: oss-prow-knative-dev,oss-prow-private-knative-dev
+    networking.gke.io/managed-certificates: oss-prow-knative-dev
 spec:
   rules:
   - host: oss-prow.knative.dev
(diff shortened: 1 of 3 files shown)


prow/oss/cluster/deck_private_deployment.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: deck-private
+  namespace: default
```
cjwagner, line 5: I'm not sure why this defaulted to `test-pods`, but it should be explicit anyways.

prow/oss/cluster/deck_private_deployment.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
```
cjwagner, line 2: This file still has some resources being applied to `test-pods`.

prow/oss/cluster/cluster.yaml
```diff
@@ -94,7 +85,7 @@ metadata:
     kubernetes.io/ingress.global-static-ip-name: "oss-prow"
     kubernetes.io/ingress.class: "gce"
     kubernetes.io/tls-acme: "true"
-    networking.gke.io/managed-certificates: oss-prow-knative-dev,oss-prow-private-knative-dev
+    networking.gke.io/managed-certificates: oss-prow-knative-dev
 spec:
   rules:
   - host: oss-prow.knative.dev
```
cjwagner, line 91: We should probably also temporarily remove the oss-prow-private.knative.dev host rule.

cjwagner (changes requested):

prow/oss/cluster/deck_private_deployment.yaml
chaodaiG: There are 2 of them, for binding deck service account with test-pods RBAC roles, not quite sure why deck need them

prow/oss/cluster/cluster.yaml
chaodaiG: removed

prow/oss/cluster/deck_private_deployment.yaml
cjwagner: I see, those are for fetching pod logs. I'm not sure if they are actually needed due to the way we use kubeconfig files, but shouldn't hurt.

cjwagner (approved):

PR #806, opened 2021-04-22
Delete GoogleCloudPlatform/k8s-cloud-provider from hmac

chaodaiG:
It's not reachable by prow, maybe google-oss-robot lost admin permission


chaodaiG:
@rramkumar1, we are modernizing how prow manages hmac tokens used for receiving github webhooks, and it requires github repo admin permission to do so. Seems like there is no prow activity in GoogleCloudPlatform/k8s-cloud-provider any more, is it still expected?

swetharepakula:
There is currently no prow activity, but we do plan on that changing. The token should not be removed.

chaodaiG:
Good to know, I'll close this PR
/close

PR #820, opened 2021-04-28
Fixing config checking jobs for private repos

chaodaiG:


prow/prowjobs/private-inrepoconfig-configcheck/chaotoppicks/great-tools/great-tools.yaml
```diff
@@ -1,6 +1,7 @@
 presubmits:
   chaotoppicks/great-tools:
   - name: pull-chaotoppicks-great-tools-validate-prow-yaml
+    clone_uri: git@github.com:chaotoppicks/great-private-tools.git
```
cjwagner, line 4: This clone URI is wrong.

prow/prowjobs/private-inrepoconfig-configcheck/chaotoppicks/great-tools/great-tools.yaml
```diff
@@ -1,6 +1,7 @@
...
     cluster: build-looker-private
```
cjwagner, line 5: It isn't safe for a public repo presubmit to run in a trusted cluster.
If this runs in a public cluster, it won't use the SSH key secret (and won't need them) so the `clone_uri` also won't be needed.

chaodaiG: Removed and switched to build cluster

cjwagner (approved):

PR #836, opened 2021-05-05
Restrict to only helltool repo from looker org

chaodaiG:
/hold
/assign @cjwagner 
cc @aryeh-looker

### DIFF
--- a/prow/oss/config.yaml
+++ b/prow/oss/config.yaml
@@ -210,7 +210,7 @@ deck:
 in_repo_config:
   enabled:
     chaotoppicks: true
-    looker: true
+    looker/helltool: true
 
   allowed_clusters:
     # TODO(chaodaiG): drop this once
@@ -218,7 +218,7 @@ in_repo_config:
     # prow
     "*": ["non-exist-cluster-since-global-cluster-is-disabled"]
     "chaotoppicks": ["build-looker-private"]
-    "looker": ["build-looker-private"]
+    "looker/helltool": ["build-looker-private"]
 
 presets:
 # docker-in-docker preset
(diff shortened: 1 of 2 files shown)


chaodaiG:
This PR enabled more plugins, but not tide (merge automation) yet. Will add tide when presubmit tests are ready

prow/oss/config.yaml
```diff
@@ -147,7 +147,7 @@ managed_webhooks:
       token_created_after: 2021-04-21T00:10:00Z
     kubeflow:
       token_created_after: 2020-12-06T00:10:00Z
-    looker:
+    looker/helltool:
```
cjwagner, line 150: Do we need to disable webhooks and `in_repo_config` for the rest of the org? That will disable automation for all other repos. I think we can keep this as is and just tweak the `in_repo_config.allowed_clusters` to limit running jobs in the private build cluster to private repos and `plugins` to limit verbose/intrusive automation to the repos we want to enable it on.

chaodaiG: My impression is that the other repos might not be ready for prow yet, so might not worth spamming `hook` with those unused webhook events. WDYT?

cjwagner: I wouldn't be worried about Hook. It can easily handle the webhook load and the plugins we configured shouldn't do anything unless they are needed with the exception of `owners-label` and `verify-owners`. Those will shallow clone the repos before seeing that there are no OWNERS files so those are the only plugins that we might want to disable org wide.

We can switch this back to org easily so this isn't a big deal. I was just thinking that it would be slightly preferable to continue supporting `/assign` and `/meow` on other repos.
/lgtm
/hold

chaodaiG: I actually agree with you. Updated to receive hook from the org

cjwagner (approved):

PR #851, opened 2021-05-18
Automated continuous deployment of oss prow

chaodaiG:
As discussed among prow oncalls, it's agreed that:
- OSS prow has been pretty stable lately
- All errors were caught by automated alerts

This gives us enough confidence to enable automated continuous deployment of oss prow, the steps are the same as https://github.com/kubernetes/enhancements/tree/master/keps/sig-testing/2539-continuously-deploy-k8s-prow

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
@@ -332,7 +332,7 @@ postsubmits:
           defaultMode: 0400
 
 periodics:
-- cron: "05 15-23 * * 1-5"  # Run at 7:05-15:05 PST (15:05 UTC) Mon-Fri
+- cron: "30 * * * 1-5"  # Bump don't label `skip-review`. Run at :30 past every hour Mon-Fri
   name: ci-oss-test-infra-autobump-prow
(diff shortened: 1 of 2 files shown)


chaodaiG:
/assign @cjwagner @fejta 
/cc @mpherman2
CC @e-blackwelder @michelle192837 @chases2

prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
```diff
@@ -366,6 +366,41 @@ periodics:
...
+  decorate: true
+  extra_refs:
+  - org: GoogleCloudPlatform
+    repo: oss-test-infra
+    base_ref: master
+  annotations:
+    testgrid-dashboards: googleoss-test-infra
+    testgrid-tab-name: autobump-prow-for-auto-deploy
+    testgrid-alert-email: k8s-infra-oncall@google.com
+    testgrid-num-failures-to-alert: '3'
```
mpherman2, line 381: This will only run once a day. I think we should be notified every time this fails. It would be bad if this failed 3 days in a row.

chaodaiG: good point, changed

fejta (approved):
/hold

PR #889, opened 2021-06-04
move bentheelder to emeritus

BenTheElder:
I have been in a different oncall and inactive here / overloaded elsewhere


BenTheElder:
there's also at least one stale entry, but keeping this PR minimal.

chaodaiG (approved):

PR #890, opened 2021-06-04
added kfp presubmit v2 test

capri-xiyue:
added kfp presubmit v2 test

### DIFF
--- a/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
+++ b/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
@@ -201,7 +201,7 @@ presubmits:
         command:
         - ./v2/test/presubmit.sh
 
-  - name: kubeflow-pipelines-v2-test
+  - name: kubeflow-pipelines-v2-go-test
     cluster: build-kubeflow
     decorate: true
     run_if_changed: "^v2/.*$"
@@ -209,7 +209,7 @@ presubmits:
       containers:
         - image: gcr.io/google.com/cloudsdktool/cloud-sdk:latest
           command:
-            - ./v2/test/presubmit-v2-tests.sh
+            - ./v2/test/presubmit-v2-go-test.sh
 
   - name: kubeflow-pipelines-manifests
     cluster: build-kubeflow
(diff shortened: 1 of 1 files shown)


capri-xiyue:
related to https://github.com/kubeflow/pipelines/pull/5802

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
```diff
@@ -200,7 +200,17 @@ presubmits:
       - image: python:3.7
         command:
         - ./v2/test/presubmit.sh
-  
+
+  - name: kubeflow-pipelines-v2-test
+    cluster: build-kubeflow
+    decorate: true
+    run_if_changed: "^(v2/.*)|(v2/test/presubmit-v2-tests.sh)$"
```
Bobgy, line 207: nit: v2/* already includes the test folder

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
```diff
@@ -200,7 +200,17 @@ presubmits:
...
+    spec:
+      containers:
+        - image: gcr.io/google.com/cloudsdktool/cloud-sdk:latest
```
Bobgy, line 210: Note, some go tests depend on a real MLMD server at localhost:8080. I think we need one more container for the MLMD server, but I've never set up a sidecar in prow tests before. We need to debug this.

Another approach can be running this test in a KFP pipeline.

capri-xiyue: Do we need side car or another KFP pipeline if we added kubectl port-fprward in the script https://github.com/kubeflow/pipelines/pull/5802/commits/57ebac924b3b9256433af1d29c5b1db660d28a75#diff-88dcf2f21baa096c2451b3d3dfbc18ad795f135bc4b260721d259345aebf1b73R35?

Bobgy: Ohh, good point! I missed the `kubectl port-forward` command.
Ideally we should test with an empty MLMD db, because some common code path will only run once, but this is good enough for now.

Issue #832, opened 2021-05-04
Migrate Prow to use Config Merger


chases2:
testgrid.k8s.io is now [using Config Merger](https://github.com/kubernetes/test-infra/pull/20986). We should use that instead of Transfigure to put our results on TestGrid, so we don't have to be gated by PR approvals.

- [x] Add Configurator job to upload config
- [x] Add validation tests to Configurator (call [Validate()](https://github.com/GoogleCloudPlatform/testgrid/blob/10af4270aad969e8ef0d349463f5108898717e53/config/config.go#L440-L441) at kubernetes/test-infra)
- [x] Add prow job to run those validation tests
- [x] Delete transfigure jobs (sync and test)
- (Note: at this point, automatic TestGrid config updates will cease until the next step is completed)
- [x] Add repository to config-merger, delete the transfigure results (at kubernetes/test-infra)
- [x] Document this process; commit this and templates (at kubernetes/test-infra)

/assign

chases2:
Configurator has a `--validate-config-file` flag. Instead of writing our own tests, the image should be run with this flag instead.

https://github.com/kubernetes/test-infra/blob/f0dcf3701878f8b9998545f5bce8fdf77f71658a/testgrid/cmd/configurator/main.go#L77

chases2:
https://github.com/kubernetes/test-infra/pull/22315 will let us just use that flag to validate Configurator's output.

PR #891, opened 2021-06-05
Remove Transfigure jobs

chases2:
Configuration is currently being uploaded directly to GCS instead.

xref #832


chases2:
I was concerned there was an issue with Config Merger, since kubevirt has already migrated and their results aren't included.

Config Merger only picks up a new configuration on startup, so the Config Merger needed to be restarted. When restarted, Kubevirt's results showed up instantly. We should have no issues adding our config also.

/hold cancel

chaodaiG (approved):

PR #905, opened 2021-06-10
Add a prow job config test checking for jobs running on hidden repos

chaodaiG:
This is generally not expected other than trusted directory where OWNERS are prow maintainers or hidden repo owners


cjwagner (approved):
Thanks Chao!

PR #911, opened 2021-06-15
Add postsubmit jobs building image for release

TAOXUY:
Our release images are built from the master head(we relied on continuous build job before) so we need a post submit job to build it.

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/esp-v2/esp-v2.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/esp-v2/esp-v2.yaml
@@ -579,10 +579,38 @@ presubmits:
         secret:
           secretName: cloudesf-testing-github-prow-service-account
 
+postsubmits:
+  GoogleCloudPlatform/esp-v2:
+  - name: ESPv2-postsubmit-build
+    cluster: espv2
+    always_run: true
+    decorate: true
+    annotations:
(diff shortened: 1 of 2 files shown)


TAOXUY:
@qiwzhang @nareddyt

nareddyt (approved):
I don't think this is needed since you can use the presubmit build when you make a PR with the changelog. But no downside to adding.

TAOXUY:
> I don't think this is needed since you can use the presubmit build when you make a PR with the changelog. But no downside to adding.

When you make another PR, it would be run on master HEAD + your changelog PR, right? No jobs run on pure master HEAD.

nareddyt:
> > I don't think this is needed since you can use the presubmit build when you make a PR with the changelog. But no downside to adding.
> 
> When you make another PR, it would be run on master HEAD + your changelog PR, right? No jobs run on pure master HEAD.

Right. Well when you make your version change PR, the build will have master HEAD + version change. That is what you want.

TAOXUY:
> > > I don't think this is needed since you can use the presubmit build when you make a PR with the changelog. But no downside to adding.
> > 
> > 
> > When you make another PR, it would be run on master HEAD + your changelog PR, right? No jobs run on pure master HEAD.
> 
> Right. Well when you make your version change PR, the build will have master HEAD + version change. That is what you want.

It is because we are doing `Squash Merge`, which will merge with a new commit and new SHA.

PR #925, opened 2021-06-22
oss-prow crier: empty global slack config and add a slack worker

chaodaiG:


prow/oss/config.yaml
```diff
@@ -228,6 +228,16 @@ deck:
...
   - looker
 
+slack_reporter_configs:
+  "*":
+    # Do not report any job types by default.
+    job_types_to_report:
+    # Do not report any job states by default.
+    job_states_to_report:
+    # By default reporting to the test channel, it is required so it has to be set here.
+    channel: channel-not-exist
```
cjwagner, line 238: I think we've done something wrong if it is actually required to set all these fields to empty or fake values. This doesn't seem consistent with the way the rest of our config works. Can you explain why an empty config would be needed here?

chaodaiG: This was a lazy copy/paste from https://github.com/GoogleCloudPlatform/oss-test-infra/blob/f29b96f29b878d7192fa3d1e72125487619f9387/prow/knative/config.yaml#L48, I assume what Chi found there still applies.

wouldn't empty channel here result in slack report error in crier?

chaodaiG:
synced offline with @cjwagner , this approach is hiding a problem in prow and we agreed that we should solve this problem from upstream first

/close

PR #942, opened 2021-07-02
Add Workload Metrics PodMonitor resources for oss-prow components.

cjwagner:
This PR should be sufficient to begin collecting metrics via Workload Metrics and it should not impact the existing monitoring solution. The `PodMonitor` resources parallel the Prometheus-Operator `ServiceMonitor` resources, but match on the pods instead of services. I've updated deployment files to name the container ports in the same way the services name them.

/assign @chaodaiG @mpherman2


chaodaiG (approved):
/hold

PR #951, opened 2021-07-07
Temporarily remove problematic PodMonitor.

cjwagner:
This PodMonitor is causing issues with the monitoring services. It has been recommended that we temporarily remove it so that we can continue the demo.
Once this PR merges I'll run `kubectl delete -n prow-monitoring podmonitor kubernetes-external-secrets`.
/assign @mpherman2 @fejta


cjwagner:
Skipping review because this is a straight forward removal of something that is already broken.

PR #954, opened 2021-07-09
Make lgtm sticky for googlers in oss-test-infra and testgrid

fejta:
/hold

### DIFF
--- a/prow/oss/plugins.yaml
+++ b/prow/oss/plugins.yaml
@@ -67,6 +67,11 @@ lgtm:
   - googleforgames
   - grpc-ecosystem
   review_acts_as_lgtm: true
+- repos:
+  - GoogleCloudPlatform/testgrid
+  - GoogleCloudPlatform/oss-test-infra
+  review_acts_as_lgtm: true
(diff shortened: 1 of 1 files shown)


michelle192837 (approved):
/hold

Unhold if you're okay with the one LGTM

prow/oss/plugins.yaml
```diff
@@ -67,6 +67,11 @@ lgtm:
   - googleforgames
   - grpc-ecosystem
   review_acts_as_lgtm: true
+- repos:
+  - GoogleCloudPlatform/testgrid
+  - GoogleCloudPlatform/oss-test-infra
+  review_acts_as_lgtm: true
+  trusted_team_for_sticky_lgtm: googlers
```
cjwagner, line 74: Using this large team will increase API token usage and potentially latency.  It should be fine for these two repos, but if the LGTM plugin becomes sluggish this is likely the culprit.

cjwagner (approved):

chases2 (approved):

PR #992, opened 2021-07-30
oss-prow reporting build clusters statuses in GCS

chaodaiG:
/hold
This won't work until version v20210730-5757a5d200, which was from https://github.com/kubernetes/test-infra/pull/23068

### DIFF
--- a/prow/oss/cluster/prow-controller-manager.yaml
+++ b/prow/oss/cluster/prow-controller-manager.yaml
@@ -87,6 +87,8 @@ kind: ServiceAccount
 metadata:
   namespace: default
   name: "prow-controller-manager"
+  annotations:
+    "iam.gke.io/gcp-service-account": "oss-prow@oss-prow.iam.gserviceaccount.com"
 ---
 kind: Role
 apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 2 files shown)


chaodaiG:
/cc @cjwagner 
cc @fejta

fejta (approved):
/hold

cjwagner (commented):
We also need to grant `prow-controller-manager` permission to upload to the GCS bucket. Specifically we need to bind its SA to a GCP SA with Workload Identity: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/d295a9af5f09cb61f17852cbcbd6bb85043acfd5/prow/oss/cluster/prow-controller-manager.yaml#L84-L89
Using the `oss-prow` SA that `crier` uses should be sufficient: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/d295a9af5f09cb61f17852cbcbd6bb85043acfd5/prow/oss/cluster/crier.yaml#L119

Before we start using this feature I'd like to resolve https://github.com/kubernetes/test-infra/pull/23072#issuecomment-890088560 since the problem encountered there seems to be from the PR that introduced this feature and we may need to revert it.

fejta (approved):

PR #995, opened 2021-08-02
oss-prow: add first stackdriver alerts via terraform

chaodaiG:


prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+locals {
+  playbook_prefix = "https://github.com/google/exposure-notifications-verification-server/blob/main/docs/playbooks/alerts"
+}
+
+resource "google_monitoring_alert_policy" "plank-alert" {
+  project      = var.project
+  display_name = "Heartbeat ProwJobs"
```
cjwagner, line 21: The display name should be consistent with the alert name, right? So I think we should name this something specific to heartbeat prowjobs rather than `plank-alert`.

chaodaiG: good point, updated

fejta (approved):
/hold

cjwagner (approved):
/hold cancel

PR #1001, opened 2021-08-04
Terraform alert on pod crashlooping

chaodaiG:
This is on top of #995 

Wasn't able to find an equivalent of `up` from prometheus https://github.com/GoogleCloudPlatform/oss-test-infra/blob/cd38a59352b0225b4ebfb5601089fa6edfa4ab83/prow/oss/cluster/monitoring/mixins/prometheus/ci_absent_alerts.libsonnet#L11, catching pod crashlooping might be a good starting point

/hold
more to be added

### DIFF
--- a/prow/oss/terraform/modules/alerts/main.tf
+++ b/prow/oss/terraform/modules/alerts/main.tf
@@ -12,6 +12,48 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+
+resource "google_monitoring_alert_policy" "pod-crashlooping" {
+  for_each     = var.prow_components
+  project      = var.project
+  display_name = "pod-crashlooping-${each.key}"
+  combiner     = "OR" # required
+
+  conditions {
+    display_name = "pod-crashlooping-${each.key}"
+
+    condition_monitoring_query_language {
+      # Alert if the service crashlooped, which results in restarts with exponential backoff.
+      # This threshold is higher than the default crashloop backoff threshold, mainly due to
+      # the fact prow components would crashloop when kubernetes master is upgrading, which
+      # normally takes 5 minutes. Setting this to 12 minutes for excluding this case
+      duration = "720s"
+      query    = <<-EOT
+      fetch k8s_container
+      | metric 'kubernetes.io/container/restart_count'
+      | filter
+          (resource.container_name == '${each.key}' && resource.namespace_name == '${each.value.namespace}')
+      | align delta(6m)
+      | every 6m
+      | group_by [], [value_restart_count_aggregate: aggregate(value.restart_count)]
+      | condition val() > 0 '1'
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
+
+  documentation {
+    content   = "The service `${each.key}` has been restarting for more than 6 minutes, very likely crashlooping."
+    mime_type = "text/markdown"
+  }
+
+  # gcloud beta monitoring channels list --project=oss-prow
+  notification_channels = ["projects/${var.project}/notificationChannels/${var.notification_channel_id}"]
+}
+
+
 resource "google_monitoring_alert_policy" "heartbeat-job-stale" {
   project      = var.project
   display_name = "heartbeat-job-stale"
--- a/prow/oss/terraform/main.tf
+++ b/prow/oss/terraform/main.tf
@@ -35,4 +35,11 @@ module "alert" {
         alert_interval = "1200s"
     }
     notification_channel_id = "14031735832803168422"
+    prow_components = {
+        "deck" = {"namespace": "default"}
+        "hook" = {"namespace": "default"}
+        "prow-controller-manager" = {"namespace": "default"}
+        "sinker" = {"namespace": "default"}
+        "tide" = {"namespace": "default"}
+    }
 }
--- a/prow/oss/terraform/modules/alerts/variables.tf
+++ b/prow/oss/terraform/modules/alerts/variables.tf
@@ -28,3 +28,10 @@ variable "heartbeat_job" {
 variable "notification_channel_id" {
   type = string
 }
+
+variable "prow_components" {
+  type = map
+  default = {
+    "svc_not_exist" = {"namespace": "default"}
+  }
+}


prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,44 @@
...
+      | condition val() > 10 '1'
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
+
+  documentation {
+    content   = "The service `${each.key}` has been restartingmore than 10 times for 10 minutes."
```
cjwagner, line 44: IIUC this query is actually checking if there are at least 10 restarts **per minute** for 10 minutes. Is that correct? If so I suspect that threshold is too high since crash loop backoff will increase the retry delay.

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,44 @@
...
+    condition_monitoring_query_language {
+      duration = "600s"
+      query    = <<-EOT
+      fetch k8s_container
+      | metric 'kubernetes.io/container/restart_count'
+      | filter
+          (resource.container_name == '${each.key}' && resource.namespace_name == '${each.value.namespace}')
+      | align rate(1m)
+      | every 1m
+      | group_by [], [row_count: row_count()]
```
cjwagner, line 34: It might be more appropriate to output the sum or restart_counts value (the number of restarts in the 1m alignment period across pods) rather than the row_count (the number of pods).

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,44 @@
...
+      | condition val() > 10 '1'
```
cjwagner, line 35: Does `val()` reference the aggregated row_count field or does this condition apply before the grouping?

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: Sorry should have mentioned, this was generated from UI, by simply selecting aggregating by `count` and interval being `1m`, not sure how those selection translated to the `group_by` line here

prow/oss/terraform/modules/alerts/main.tf
cjwagner: This seems wrong to me then then. Count is the number of timeseries results, not the number of restarts which is what we should be conditioning on. Rather than selecting aggregating the `count` in the UI, was there an option for aggregating the `sum` instead?

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,44 @@
...
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
+
+  documentation {
+    content   = "The service `${each.key}` has been restarting more than 10 times within 30 minutes."
```
cjwagner, line 44: I think this is still inaccurate. Now were checking if there are 10 restarts per every 5mins for a 30min time span. This is a bit confusing, is a long alerting delay, and may still not fire due to crash loop backoff.  How about alerting if a service restarts more than 3 times in 5 mins?

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,44 @@
...
+    display_name = "ci-absent-${each.key}"
+
+    condition_monitoring_query_language {
+      duration = "1800s"
+      query    = <<-EOT
+      fetch k8s_container
+      | metric 'kubernetes.io/container/restart_count'
+      | filter
+          (resource.container_name == '${each.key}' && resource.namespace_name == '${each.value.namespace}')
+      | align rate(5m)
```
cjwagner, line 32: Rate is going to return a value per second rather than the increase in value over the time period. e.g if there are 5 restarts in the last 5mins this will return 1/60, not 5.

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: I think you're right. count of rows doesn't reflect how many times a pod restarts

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: 3 times in 5 mins will cause false positive, especially when cluster master is upgrading, some components would crashlooping for 5-10 minutes

chaodaiG:
> Rate is going to return a value per second rather than the increase in value over the time period. e.g if there are 5 restarts in the last 5mins this will return 1/60, not 5.

Agreed. I'll update the condition

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: updated with `delta` aligner

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: ack, moved away from using rate

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,48 @@
...
+      | condition val() > 5 '1'
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
+
+  documentation {
+    content   = "The service `${each.key}` has been restarting more than 5 times for 10 minutes, very likely crashlooping."
```
cjwagner, line 48:
```suggestion
    content   = "The service `${each.key}` has been restarting more than 5 times per 5 minutes for 10 minutes, very likely crashlooping."
```

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,48 @@
...
+  for_each     = var.prow_components
+  project      = var.project
+  display_name = "pod-crashlooping-${each.key}"
+  combiner     = "OR" # required
+
+  conditions {
+    display_name = "pod-crashlooping-${each.key}"
+
+    condition_monitoring_query_language {
+      # Alert if the service crashlooped, which results in more than 5 restarts in 5 minutes.
```
cjwagner, line 26: These [k8s docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-policy) indicate the crash loop backoff delay is used and increased after the first restart. The delay can reach 5 mins in which case we'd only get 1 restart per 5 mins.  So IIUC it would be impossible for this alert's condition to be met.

chaodaiG: I think you're right. I did see that doc but didn't compute it clearly.

event: restart, 10s ,restart,  20s ,restart,  40s ,restart,  80s ,restart,  160s ,restart,  300s ,restart

| time(s) | total restart on real crashlooping | restart in 10 minutes| total restart on master restart | restart in 10 minutes|
|-------|-----------|-------------|------------|--------------|
|0|1|1|1|1|
|10|2|2|2|2|
|30|3|3|3|3|
|70|4|4|4|4|
|150|5|5|5|5|
|310|6|5|6|5|
|330|6|4|6|4|
|370|6|3|6|3|
|450|6|2|6|2|
|610|7|2|6|1|
|630|7|1|6|0|
|670|7|1|6|0|
|750|7|1|6|0|
|910|7|1|6|0|

so in master restart case, there will be 1 restart in 10 minutes (0s to 610s), and 2 restarts in 440s(10s to 450s).
in real crashlooping case, there will be 2 restart in 10 minutes (0s to 610s), this might not be guaranteed depends on the resolution of the data and the cutoff of intervals.

I think it might be good to do either 1 restart in 12 minutes, or 2 restarts in 8 minutes. WDYT?

cjwagner: We don't want to alert on a single restart. Maybe check for at least 1 restart per 6 minutes (checked every 6 minutes) for at least 12 minutes.  That would require at least 2 restarts in total so we'd avoid transient errors, but we shouldn't fire on any number of restarts within a 6 minute window as long as they are contained to that window.

chaodaiG: agreed, done

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,48 @@
...
+      # This threshold is higher than the default crashloop backoff threshold, mainly due to
+      # the fact prow components would crashloop when kubernetes master is upgrading, which
+      # normally takes 5 minutes. Setting this to 10 minutes for excluding this case
+      duration = "1200s"
```
cjwagner, line 30: This is 20mins not 12mins. It might be clearer to use `m` units rather than `s` if that is supported. The comment also needs to be updated.

chaodaiG: unfortunately `m` is not supported, updated the rest

cjwagner (approved):

PR #1004, opened 2021-08-05
Add podmonitor for external secrets and sub

chaodaiG:
Noticed that the metrics port was not added initially, which probably was the reason why podmonitor was not so happy last time

### DIFF
--- a/prow/oss/cluster/monitoring/prow_podmonitors.yaml
+++ b/prow/oss/cluster/monitoring/prow_podmonitors.yaml
@@ -151,3 +151,42 @@ spec:
   selector:
     matchLabels:
       app: crier
+---
+apiVersion: monitoring.gke.io/v1alpha1
+kind: PodMonitor
+metadata:
+  labels:
+    app: sub
+  name: sub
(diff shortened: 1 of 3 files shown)


prow/oss/cluster/monitoring/prow_podmonitors.yaml
```diff
@@ -151,3 +151,42 @@ spec:
...
+metadata:
+  labels:
+    app.kubernetes.io/name: kubernetes-external-secrets
+    app: kubernetes-external-secrets
+  name: kubernetes-external-secrets
+  namespace: prow-monitoring
+spec:
+  podMetricsEndpoints:
+  - interval: 30s
+    port: prometheus
```
cjwagner, line 185: > Noticed that the metrics port was not added initially, which probably was the reason why podmonitor was not so happy last time

The pod monitor uses the `prometheus` port. I don't think 9090 is a valid port number for KES.

cjwagner (changes requested):

chaodaiG:
/close
will reopen when it's ready

PR #1016, opened 2021-08-10
Add terraform config for a predicted-gh-rate-limit-exhaustion alert.

cjwagner:
/assign @chaodaiG

### DIFF
--- a/prow/oss/terraform/modules/alerts/main.tf
+++ b/prow/oss/terraform/modules/alerts/main.tf
@@ -12,6 +12,58 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+resource "google_monitoring_alert_policy" "predicted-gh-rate-limit-exhaustion" {
+  project      = var.project
+  display_name = "predicted-gh-rate-limit-exhaustion"
+  combiner     = "OR" # required
+
+  conditions {
+    display_name = "predicted-gh-rate-limit-exhaustion"
+
+    condition_monitoring_query_language {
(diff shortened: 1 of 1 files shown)


fejta (approved):
/hold

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -12,6 +12,58 @@
...
+      | filter val() < 250
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
+
+  documentation {
+    content   = "One of the GitHub tokens used with `ghproxy` is predicted to exhaust its rate limit before the end of the rate limit reset window."
```
cjwagner, line 60: It would be handy to embed contextual information from the alerts into these messages. e.g. The actual number of predicted tokens remaining at the end of the reset window or the GH login of the token that the alert is firing for (via a label). I see examples below that use variables from terraform, but I'd like to use values from the MQL query results. Do you know if this is possible, or how to do it? The docs for the `documentation` field don't mention this: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy#documentation

cjwagner (commented):
Is there a way to validate these files? It would be nice to do so locally and/or in presubmit.

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: I haven't seen it either, not sure whether it's supported by cloud monitoring or not

chaodaiG (approved):

PR #1020, opened 2021-08-10
Terraform: http uptime check alert

chaodaiG:


prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -133,3 +133,37 @@ resource "google_monitoring_alert_policy" "heartbeat-job-stale" {
...
+      | condition val() > 1 '1'
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
+
+  documentation {
+    content   = "Host Down"
```
cjwagner, line 163: This is not going to be enough information to address the alert once we have more than one host being monitored. We need to include the URL of the host that is down somehow.

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -133,3 +133,37 @@ resource "google_monitoring_alert_policy" "heartbeat-job-stale" {
...
+  display_name = "HostDown"
+  combiner     = "OR"
+  conditions {
+    display_name = "Host is unreachable"
+    condition_monitoring_query_language {
+      duration = "120s"
+      query    = <<-EOT
+      fetch uptime_url
+      | metric 'monitoring.googleapis.com/uptime_check/check_passed'
+      | filter (metric.check_id == 'blackbox-exp')
```
cjwagner, line 149: What is this field and where does its value come from? I only see this value in the existing metrics.

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -133,3 +133,37 @@ resource "google_monitoring_alert_policy" "heartbeat-job-stale" {
...
+      | align next_older(1m)
```
cjwagner, line 150: The choice between `next_older` and `next_younger` doesn't matter here, right? You're just trying to grab a discrete value without actual aggregation?

prow/oss/terraform/main.tf
```diff
@@ -42,4 +42,7 @@ module "alert" {
         "sinker" = {"namespace": "default"}
         "tide" = {"namespace": "default"}
     }
+    blackbox_probers = [
+        "oss-prow.knative.dev"
```
cjwagner, line 46: I see an existing uptime check for `testgrid.k8s.io`, was that manually added for testing?

cjwagner (approved):
/hold
I just have nits

prow/oss/terraform/main.tf
chaodaiG: yep

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: > The choice between `next_older` and `next_younger` doesn't matter here, right? You're just trying to grab a discrete value without actual aggregation?

this was adapted from auto-generated alerts from UI, where we can set an alert from `uptime check` page

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: this was from the experiment, removed

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: Unfortunately this is not available, but I believe that the alert email will tell us more

cjwagner:
LGTM. Ready to `/unhold` this? (`lgtm` is sticky)

PR #1025, opened 2021-08-11
webhook missing alert with terraform

chaodaiG:


prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -141,3 +141,40 @@ resource "google_monitoring_alert_policy" "heartbeat-job-stale" {
...
+      | condition val(0) == 0 && is_business_hour * (1-is_weekend) == 1
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
+
+  documentation {
+    content   = "Webhook missing."
```
cjwagner, line 174: nit: It would be good to mention for how long so that oncall can get a feel for severity.

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -141,3 +141,40 @@ resource "google_monitoring_alert_policy" "heartbeat-job-stale" {
...
+  conditions {
+    display_name = "webhook-missing"
+
+    condition_monitoring_query_language {
+      duration = "720s"
+      query    = <<-EOT
+      fetch k8s_container
+      | metric 'workload.googleapis.com/prow_webhook_counter'
+      | sum
+      | align delta_gauge()
```
cjwagner, line 159: nit: I don't think this needs to be converted to a gauge, what is the motivation?

cjwagner (approved):
/hold

prow/oss/terraform/modules/alerts/main.tf
chaodaiG: when I uses `delta` on UI, the json generated at the last line is something like `val() == 0 'no_unit_value'`, changing this to gauge just for avoiding the annoyance

prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -166,9 +166,47 @@ resource "google_monitoring_alert_policy" "probers" {
...
+      EOT
+      trigger {
+        count = 1
+      }
+    }
+  }
 
   documentation {
-    content   = "Host Down"
+    content   = "Webhook missing for 20 minutes."
```
cjwagner, line 209: The duration is set to 720s which is 12mins, not 20mins.

chaodaiG: what a math, lol

chaodaiG: 20 minutes is what I want, fixed above

cjwagner (approved):

PR #1026, opened 2021-08-11
Fix heartbeat job alert by summing values from all pods

chaodaiG:


prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -123,9 +123,10 @@ resource "google_monitoring_alert_policy" "heartbeat-job-stale" {
       | filter
           (metric.job_name == '${var.heartbeat_job.job_name}'
           && metric.state == 'success')
-      | align delta(${var.heartbeat_job.interval})
+      | sum # Combining values reported by all prow-controller-manager pods
+      | align delta_gauge(${var.heartbeat_job.interval})
```
cjwagner, line 127: Probably doesn't make a difference here, but I don't think this needs to be converted to a gauge.

cjwagner (approved):

PR #1034, opened 2021-08-13
Add boskos alerts

chaodaiG:
boskos doesn't exist in oss-prow, this module can be imported by other projects for monitoring boskos resources

### DIFF
--- a/prow/oss/terraform/modules/alerts/boskos/main.tf
+++ b/prow/oss/terraform/modules/alerts/boskos/main.tf
@@ -1,0 +1,97 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Enabled only when allowed_list is empty, this is default behavior
+resource "google_monitoring_alert_policy" "boskos-alerts" {
+  count        = length(var.allowed_list) == 0 ? 1 : 0
+  project      = var.project
+  display_name = "boskos-alerts"
+  combiner     = "OR" # required
+
+  conditions {
+    display_name = "Boskos ran out of resources"
+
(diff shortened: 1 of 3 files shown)


prow/oss/terraform/modules/alerts/boskos/main.tf
```diff
@@ -0,0 +1,53 @@
...
+      fetch k8s_container
+      | metric 'workload.googleapis.com/boskos_resources'
+      | {
+          t_0:
+          filter state == 'free'
+          ;
+          t_1:
+          ident
+      }
+      | group_by [metric.type, metric.instance]
```
cjwagner, line 36: Grouping by instance might not be valid or make sense any more. We'll probably want to group by project and cluster instead?

cjwagner, line 36: I think `sum` is the default aggregator, but it might be clearer to explicitly `sum(val())`.

prow/oss/terraform/modules/alerts/boskos/main.tf
```diff
@@ -0,0 +1,53 @@
...
+      | outer_join 0
+      | condition t_0.value_prowjob_state_transitions_aggregate == 0 && t_1.value_prowjob_state_transitions_aggregate > 5
```
cjwagner, line 38: These field names look like they need to be updated.

prow/oss/terraform/modules/alerts/boskos/main.tf
chaodaiG: agreed that instance doesn't make sense any more, removed. I think `project` and `cluster` would be relevant if alerting on boskos metrics from other projects

chaodaiG: I have tried `sum(val())` in UI, it doesn't work, `sum` worked but aggregated values from all types, which is not what we want

prow/oss/terraform/modules/alerts/boskos/main.tf
chaodaiG: These fields names were intellisensed from UI, they worked and look like `sum` to me, but I haven't found any documentation about this yet

cjwagner: Are you sure they're the right metric? We should be working with boskos resource counts, not prowjob state transitions.

cjwagner (commented):
I agree that this will be easier to set up once we have real data to work with. I'll work on that.

chaodaiG:
Have updated based on experiments done with prow instance that has boskos metrics, and I think it works now

chaodaiG:
@cjwagner please take a look when you got a chance

PR #1044, opened 2021-08-16
Increase hook replicas to 6 and set resource requests

chaodaiG:
Hook is responsible for triggering inrepoconfig jobs, it figures out how to run it by cloning the entire target repo, which could be throttled when encounters large repo at scale. Increasing the replicas should be able to help


cjwagner (approved):
Thanks, Chao!

PR #1045, opened 2021-08-16
Add liveliness and readinessprobe for hook

chaodaiG:
For consistency reason with other prow instances, also makes the service more resilient


chaodaiG:
fixed merge conflict, may I have another lgtm please?

fejta (approved):

fejta:
> fixed merge conflict, may I have another lgtm please?

LGTM should be sticky for us, aka not removed after we get it

PR #1052, opened 2021-08-18
Jobs in test-infra-trusted cluster use WI by default

chaodaiG:
/hold


chaodaiG:
The SA and WI were both set up
/unhold

michelle192837 (approved):

PR #1079, opened 2021-08-30
Switch Deck Private instance from hidden only to tenantID

mpherman2:
/assign @cjwagner @chaodaiG 
/hold


chaodaiG:
Great to see!

can we evaluate this quickly by setting `tenant-id` on default deck and see what's going on?

chaodaiG (approved):
Had an offline discussion with @mpherman2 , we agree that it's safe to do this first

PR #1088, opened 2021-09-02
Update rbac.authorization.k8s.io to v1

chaodaiG:
v1beta1 is deprecated v.1.22+ as https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122, updating it before we ran into it


fejta (approved):
/hold

fejta (approved):

PR #1098, opened 2021-09-07
Switch Deck Private instance from hidden only to tenantID

mpherman2:
We fixed the issue with labeling Prowjobs and created some integration tests to confirm this works as expected.
/hold
/assign @chaodaiG @cjwagner


chaodaiG (approved):
/hold

PR #1115, opened 2021-09-14
oss-prow: move crier and tide to dedicated nodepool

chaodaiG:
These are components that might experience high cpu and disk I/O for inrepoconfig lookup, move them to a dedicated nodepool, so that they can be scheduled on a node with no neighbor, and the availability is guaranteened


chaodaiG:
/hold
the nodepool is not created yet

PR #1170, opened 2021-10-06
Add reporting template for looker/helltool pull requests

vanjavucinic-looker:


prow/oss/config.yaml
```diff
@@ -84,6 +84,8 @@ plank:
   pod_pending_timeout: 15m
   pod_unscheduled_timeout: 5m
   build_cluster_status_file: 'gs://oss-prow/oss-prow-build-clusters-statuses.json'
+  report_templates:
+    "looker/helltool": "View Performance Test Results Here:\n  FrontEnd Tests: https://meta.looker.com/dashboards/8194?My%20Commit={{.Spec.Refs.Pulls[0].SHA}}\n  Backend Tests: https://meta.looker.com/dashboards/8195?My%20Commit={{.Spec.Refs.Pulls[0].SHA}}"
```
listx, line 88: Looks like you cannot index with `[`. Instead I would suggest you do something like `{{with index .Spec.Refs.Pulls 0}}` as per https://github.com/GoogleCloudPlatform/oss-test-infra/blob/645926b1d9e01230c9aa575d41ba87068e52f23e/prow/knative/config.yaml#L18

vanjavucinic-looker: Changed, thanks :)

PR #1179, opened 2021-10-08
Increase ratelimit exhaustion prediction alert alignment window from 10m to 15m.

cjwagner:
This will make the alerts less sensitive to sudden increases in consumption at the cost of alerting later.
/assign @listx @mpherman2


cjwagner:
Applied.

PR #1183, opened 2021-10-11
Fix ingress for /ghapp-hook/hook

chaodaiG:
The existing ingress didn't work, very likely due to two ingresses using the same static ip address

### DIFF
--- a/prow/oss/cluster/cluster.yaml
+++ b/prow/oss/cluster/cluster.yaml
@@ -104,38 +104,7 @@ metadata:
     kubernetes.io/ingress.class: "gce"
     kubernetes.io/tls-acme: "true"
     networking.gke.io/managed-certificates: oss-prow-knative-dev,oss-prow-private-knative-dev
-spec:
-  rules:
-  - host: oss-prow.knative.dev
-    http:
-      paths:
-      - path: /*
-        backend:
-          serviceName: deck
-          servicePort: 80
-      - path: /hook
-        backend:
-          serviceName: hook
-          servicePort: 8888
-  - host: oss-prow-private.knative.dev
-    http:
-      paths:
-      - path: /*
-        backend:
-          serviceName: deck-private
-          servicePort: 80
----
-apiVersion: extensions/v1beta1
-kind: Ingress
-metadata:
-  name: hook-ghapp-ing
-  namespace: default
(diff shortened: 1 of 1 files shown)


chaodaiG:
As I read more into this, I might have misunderstood how rewrite-target works, which is not supported by gce ingress according to https://github.com/kubernetes/ingress-gce/issues/109. If we switch to nginx ingress we'll need to manually create the Loadbalancer for binding static ip

chaodaiG:
According to https://cloud.google.com/kubernetes-engine/docs/how-to/managed-certs#prerequisites, using Google managed cert will also not work with nginx ingress

chaodaiG:
Depends on https://github.com/kubernetes/test-infra/pull/23972

prow/oss/cluster/hook-ghapp.yaml
```diff
@@ -28,6 +28,7 @@ spec:
         image: gcr.io/k8s-prow/hook:v20211008-562307626f
```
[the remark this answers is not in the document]
chaodaiG: good point, was thinking about two PRs, doesn't seem to be necessary

prow/oss/cluster/cluster.yaml
```diff
@@ -117,42 +117,20 @@ spec:
...
           servicePort: 8888
+      - path: /ghapp-hook/hook
+        backend:
+          serviceName: hook-ghapp
+          servicePort: 8888
   - host: oss-prow-private.knative.dev
     http:
       paths:
-      - path: /*
+      - path: (/.*)
```
cjwagner, line 127: This needs to be kept as `/*`

prow/oss/cluster/hook-ghapp.yaml
```diff
@@ -28,6 +28,7 @@ spec:
...
         imagePullPolicy: Always
         args:
+        - --webhook-path=/ghapp-hook/hook
```
cjwagner, line 31: nit: The `/hook` suffix is no longer needed. We could just use `/ghapp-hook` here and in the ingress rule.

chaodaiG: done

cjwagner (approved):

PR #1184, opened 2021-10-11
Opt in looker for GitHub app authentication

chaodaiG:
Looker had already installed the app https://github.com/apps/google-oss-prow, so this PR would work out of box.

Note that there will be some minor impacts:
1. there will be a few seconds that webhooks are double handled, so prow might do something repeatedly
2. prow will not recognize the comments it left before, so on existing PRs there will be two prow comments instead of one


chaodaiG:
/assign @cjwagner 
cc @2nd-half

cjwagner (approved):
/hold

cjwagner (approved):
/hold

chaodaiG:
/unhold
🤞

PR #1202, opened 2021-10-14
Explicitly opt-in GCP repos to plugins that consume ratelimit or act passively.

cjwagner:
/assign @chaodaiG 
Hold to add unit tests to prevent additional plugins from being enabled at org scope.
/hold


chaodaiG (approved):

cjwagner:
> Hold to add unit tests to prevent additional plugins from being enabled at org scope.

Ugh, I'm running into lots of problems with go version and go mod dependencies. I'm giving up for now, could probably use a hand tomorrow.

fejta (approved):
/hold

PR #1236, opened 2021-10-27
Branch protect only the branches we care about

krsna-m:
Fixes: https://github.com/knative/test-infra/issues/2859
Related: https://github.com/knative/test-infra/issues/2929

### DIFF
--- a/prow/knative/config.yaml
+++ b/prow/knative/config.yaml
@@ -111,7 +111,7 @@ deck:
     - lens:
         name: junit
       required_files:
-      # Nested folders is a valid use case  
+      # Nested folders is a valid use case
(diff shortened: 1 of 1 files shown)


prow/knative/config.yaml
```diff
@@ -200,6 +200,10 @@ branch-protection:
         - tide
       exclude:
       - "^dependabot/" # don't protect branches created by dependabot
+      include:
+      - "^main$"
+      - "^master$"
+      - "^release-*"
```
dprotaso, line 206: I think the above regex is missing a `.` 

```suggestion
      - "^release-.*"
```

Issue #1171, opened 2021-10-06
OSS prow uses GitHub app for authenticating


chaodaiG:
OSS prow currently uses bot personal access token(PAT) for authentication with GitHub APIs, which has a global rate limit of 5000 per hour. This is not very well scalable with many tenants. Switching over to GitHub app will get a rate limit per installation, which would greatly increase the rate limit exhaustion problem we have seen lately

chaodaiG:
The migration can be done one org/repo at a time, there are two steps:
1. Install Google OSS Prow app on corresponding org/repo https://github.com/apps/google-oss-prow
2. Register the org/repo to be handled by prow deployments that use GH app for authentication, example PR #1184 

The org/repo with [managed webhooks](https://github.com/GoogleCloudPlatform/oss-test-infra/blob/e722a2f746cb6e09589e5227ffef2713c1a8e1d0/prow/oss/config.yaml#L202) are the targets for migration:

- [ ] google/http_pattern_matcher
- [ ] googleforgames/agones
- [ ] grpc-ecosystem/grpc-httpjson-transcoding
- [ ] GoogleCloudPlatform/artifact-registry-apt-transport
- [ ] GoogleCloudPlatform/artifact-registry-yum-plugin
- [ ] GoogleCloudPlatform/compute-image-tools
- [ ] GoogleCloudPlatform/elcarro-oracle-operator
- [ ] GoogleCloudPlatform/esp-v2
- [ ] GoogleCloudPlatform/guest-agent
- [ ] GoogleCloudPlatform/guest-configs
- [ ] GoogleCloudPlatform/guest-diskexpand
- [ ] GoogleCloudPlatform/guest-logging-go
- [ ] GoogleCloudPlatform/guest-oslogin
- [ ] GoogleCloudPlatform/guest-test-infra
- [ ] GoogleCloudPlatform/k8s-cloud-provider
- [ ] GoogleCloudPlatform/osconfig
- [ ] GoogleCloudPlatform/oss-test-infra
- [ ] GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp
- [ ] GoogleCloudPlatform/testgrid
- [ ] kubeflow
- [ ] looker
- [ ] GoogleCloudPlatform/blueprints
- [ ] chaotoppicks

Note:
[...]

PR #1246, opened 2021-10-29
testgrid repo uses github app for authentication

chaodaiG:
This is part of https://github.com/GoogleCloudPlatform/oss-test-infra/issues/1171, I have requested installation of all repos under GoogleCloudPlatform that are currently using prow, and so far it seems like they are all approved. Trying this out on testgrid repo, will migrate others all at once if this proves to be working

### DIFF
--- a/prow/oss/cluster/crier-ghapp.yaml
+++ b/prow/oss/cluster/crier-ghapp.yaml
@@ -38,6 +38,7 @@ spec:
         - --github-enabled-org=chaotoppicks
         - --github-enabled-org=looker
         - --github-enabled-repo=GoogleCloudPlatform/oss-test-infra
+        - --github-enabled-repo=GoogleCloudPlatform/testgrid
         - --github-workers=6
         - --job-config-path=/etc/job-config
         - --kubernetes-blob-storage-workers=1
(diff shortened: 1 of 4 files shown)


prow/oss/cluster/crier-ghapp.yaml
```diff
@@ -38,6 +38,7 @@ spec:
         - --github-enabled-org=chaotoppicks
         - --github-enabled-org=looker
         - --github-enabled-repo=GoogleCloudPlatform/oss-test-infra
+        - --github-enabled-repo=GoogleCloudPlatform/testgrid
```
cjwagner, line 41: Don't we also need to disable on the bot based `hook` and `crier` instances?

cjwagner (changes requested):

prow/oss/cluster/crier-ghapp.yaml
chaodaiG: good catch

chaodaiG: fixed

michelle192837 (approved):

PR #1247, opened 2021-10-29
Make helltool checkconfig job include core job config.

cjwagner:
I don't think this will make a difference, but should be done for consistency. May be important if we start caring about naming collisions later or for some subtle reasons.
/assign @chaodaiG 
/hold


chaodaiG (approved):

cjwagner:
I confirmed the job is still passing.

PR #1249, opened 2021-10-29
GoogleCloudPlatform all switch over to use GitHub app

chaodaiG:
#1246 had been proved to be working, switching all over now is not super risky

/hold


chaodaiG:
will unhold on Monday

PR #1260, opened 2021-11-02
Make OSS Prow config validation check for unknown fields in job config.

cjwagner:
/assign @listx

### DIFF
--- a/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-postsubmits.yaml
+++ b/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-postsubmits.yaml
@@ -5,21 +5,6 @@ postsubmits:
     branches:
     - ^master|release-.+$
     decorate: true
-    env:
-    - name: DOCKER_IN_DOCKER_ENABLED
-      value: "true"
-    volumes:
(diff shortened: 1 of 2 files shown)


prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-postsubmits.yaml
```diff
@@ -5,21 +5,6 @@ postsubmits:
...
-    - name: docker-graph
-      emptyDir: {}
-    # krte (normal) path
-    - name: docker-root
-      emptyDir: {}
-    volumeMounts:
-    - name: docker-graph
-      mountPath: /docker-graph
-    - name: docker-root
-      mountPath: /var/lib/docker
```
cjwagner, line 22: @Bobgy FYI, these fields were at the wrong indentation level and therefore were not actually applied. I'm removing them to avoid changing the job behavior, but I wanted to let you know in case you'd like to try adding the fields in the correct place. 
Once this PR merges we'll have presubmit validation to catch this type of problem.

Bobgy: Thank you for the heads up!

PR #1267, opened 2021-11-04
Update uptime probe alerts to ignore extraneous uptime_checks.

cjwagner:
Without this, `uptime_checks` configured in scoped projects will also be included, but we only care about the `uptime_checks` we are explicitly enumerating here.

/assign @chaodaiG

### DIFF
--- a/prow/oss/terraform/modules/alerts/main.tf
+++ b/prow/oss/terraform/modules/alerts/main.tf
@@ -212,6 +212,7 @@ resource "google_monitoring_alert_policy" "probers" {
       fetch uptime_url
       | metric 'monitoring.googleapis.com/uptime_check/check_passed'
       | align next_older(1m)
+      | filter resource.project_id == '${var.project}'
       | every 1m
       | group_by [resource.host],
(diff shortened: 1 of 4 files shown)


prow/oss/terraform/modules/alerts/main.tf
```diff
@@ -212,6 +212,7 @@ resource "google_monitoring_alert_policy" "probers" {
       fetch uptime_url
       | metric 'monitoring.googleapis.com/uptime_check/check_passed'
       | align next_older(1m)
+      | filter resource.project_id == '${var.project}'
```
chaodaiG, line 215: are we sure `project_id` is a reliable differentiator?

chaodaiG (approved):
/hold
see my comment

prow/oss/terraform/modules/alerts/main.tf
cjwagner: Yes it is. Previously I was setting the value based on the `blackbox_prober` map values. Now they're all set to a consistent value (the metrics project).

cjwagner (commented):
/hold cancel

PR #1277, opened 2021-11-05
github/google org switches to use GitHub app for prow

chaodaiG:
The prow app was installed in http_pattern_matcher and bms-toolkit repos


chaodaiG:
cc @nareddyt for http_pattern_matcher
cc @jcnars for bms-toolkit

PR #1281, opened 2021-11-08
Use updated kubeconfig which doesn't have build-csi cluster

chaodaiG:


chaodaiG:
I have updated the kubeconfig secret 
/hold
I'll monitor once this merges

michelle192837 (approved):

PR #1291, opened 2021-11-10
Remove all managed webhooks

chaodaiG:
Now that no oss-prow tenant uses PAT, there is no more webhooks to be managed, removing entirely to avoid confusion. This change doesn't affect webhooks set up by GitHub app

### DIFF
--- a/prow/oss/config.yaml
+++ b/prow/oss/config.yaml
@@ -205,16 +205,6 @@ prowjob_default_entries:
     config:
       tenant_id: 'blueprints'
 
-managed_webhooks:
-  # Will accept invitations from org/repo listed under here
-  auto_accept_invitation: true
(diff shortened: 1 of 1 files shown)


chaodaiG:
/hold
Better to be after #1290 and #1287

cjwagner (commented):
Please also remove corresponding the HMAC reconciler postsubmit job.

chaodaiG:
Will need that to remove the remaining HMAC tokens as well as webhooks when this PR is merged. I'll create a PR deleting the job once that's done

cjwagner:
FWIW the postsubmit will run after this PR merges even if the PR removes it. That would just be the last run. Removing in a separate PR also works though.

cjwagner (approved):

chaodaiG:
/unhold
Just in case the postsubmit job flakes

PR #1292, opened 2021-11-10
Migrate tide to use GitHub app

chaodaiG:
Now that all prow tenants have migrated to use GitHub apps, migrate tide to use app authentication should be no blockers.

Note that tide doesn't get the tokens benefits as hook and crier, so we might still run into rate limit issue when there are lots of repos enable tide. However it shouldn't be worse than today

### DIFF
--- a/prow/oss/cluster/tide.yaml
+++ b/prow/oss/cluster/tide.yaml
@@ -26,34 +26,41 @@ spec:
         - --dry-run=false
         - --history-uri=gs://oss-prow/tide-history.json
         - --status-path=gs://oss-prow/tide-status-checkpoint.yaml
-        - --github-token-path=/etc/github/oauth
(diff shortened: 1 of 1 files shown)


chaodaiG:
/hold
Will unhold when I have time to monitor

cjwagner (approved):
> Note that tide doesn't get the tokens benefits as hook and crier, so we might still run into rate limit issue when there are lots of repos enable tide. However it shouldn't be worse than today

Yeah, I don't expect this to be a problem, Tide is relatively rate limit efficient. The biggest cost is setting statuses on PRs, but that is technically best effort and is already rate limit throttled.

chaodaiG:
tide was tested in https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1293, it worked

PR #1293, opened 2021-11-10
Migrate sub to use GitHub app

chaodaiG:
Now that all prow tenants have migrated to use GitHub apps, migrate sub to use app authentication should be no blockers.

The only place where GitHub authentication is used in sub is for cloning repos to figure out inrepoconfig prow jobs, this had been proved to be working in hook, the risk should be low


cjwagner (approved):

chaodaiG:
/unhold
by the way testing tide

PR #1295, opened 2021-11-10
Oss prow: crier and hook use GitHub app only

chaodaiG:
Now that everything has migrated, crier and hook deployments that use PAT are no longer serving any traffics. Remove them so that we don't need to do the --enabled-repos and --disabled-repos trick any more


chaodaiG:
/hold
Will unhold when I have time to monitor

fejta (approved):
/hold

PR #1299, opened 2021-11-11
Revert "Oss prow: crier and hook use GitHub app only"

chaodaiG:
This reverts commit 063b0e4a405ccc0816a05560b0dbf5faca3c3771.

Hook doesn't seem to work, will debug


chaodaiG:
/hold
would prefer #1300 over this one

chaodaiG:
it should be fixed now, if not this command would not work:
/close

PR #1309, opened 2021-11-16
compute-image-tools: Add presubmit for module tests

EricEdens:
This configures a new job that will run the module tests in `cli_tools_test/module`. They're configured using the recommendation from  [jobs.md](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md):

> A useful pattern when adding new jobs is to start with always_run set to false and skip_report set to true. Test it out a few times by manually triggering, then switch always_run to true. Watch for a couple days, then switch skip_report to false.

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
@@ -1,3 +1,31 @@
+postsubmits:
+  GoogleCloudPlatform/compute-image-tools:
+  - name: cli-tools-postsubmit-module-tests
+    cluster: gcp-guest
+    decorate: true
+    annotations:
+      testgrid-dashboards: google-gce-compute-image-tools
+      testgrid-tab-name: module-tests
+      description: "Postsubmit tests; executes cli_tools_tests/module/"
+    branches:
+      - ^master$
+    labels:
+      preset-daisy: "true"
+    spec:
+      containers:
+        - image: gcr.io/gcp-guest/gotest:latest
+          imagePullPolicy: Always
+          command:
+            - "/go/main.sh"
+          args:
+           - "cli_tools_tests/module/"
+           - "-test.timeout=30m"
+          env:
+            - name: GOOGLE_CLOUD_PROJECT
+              value: compute-image-test-pool-001
+            - name: GOOGLE_CLOUD_ZONE
+              value: us-central1-a
+
 presubmits:
   GoogleCloudPlatform/compute-image-tools:
   - name: compute-image-tools-flake8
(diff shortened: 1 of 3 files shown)


prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
```diff
@@ -41,6 +41,24 @@ presubmits:
         command:
         - "/go/main.sh"
         args: ["cli_tools/"]
+  - name: cli-tools-presubmit-module-tests
+    cluster: gcp-guest
+    skip_report: true
```
hopkiw, line 46: you won't see the result with this option. i think there is a better testing strategy than committing then separately enabling, and you should just submit the final enabled content, here. sync with me offline on the testing strategy

prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
```diff
@@ -41,6 +41,24 @@ presubmits:
...
+    always_run: false
+    run_if_changed: 'cli_tools/.*'
+    trigger: "(?m)^/cli-tools-module-tests$"
+    rerun_command: "/cli-tools-module-tests"
+    context: prow/presubmit/module-tests
+    decorate: true
+    labels:
+      preset-daisy: "true"
```
hopkiw, line 54: the gotest container doesn't need the daisy credentials, does it?

hopkiw (changes requested):

prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
EricEdens: The tests in `cli_tools_tests/module/` execute using the go test framework, but use the credentials while they run.

hopkiw: per our offline discussion, once you've done testing, you should add a comment here explaining that this is an integ test and needs creds, and add any other necessary environment variables to the job spec

prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
```diff
@@ -1,3 +1,32 @@
+postsubmits:
```
EricEdens, line 1: Switched to postsubmit since I found that running as a presubmit took too long (~30 minutes)

prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
```diff
@@ -1,3 +1,32 @@
...
+      preset-daisy: "true"
+    spec:
+      containers:
+        - image: gcr.io/gcp-guest/gotest:latest
+          imagePullPolicy: Always
+          command:
+            - "/go/main.sh"
+          args:
+           - "cli_tools_tests/module/"
+           - "-test.timeout=30m"
```
EricEdens, line 23: Added support for extra test args here: https://github.com/GoogleCloudPlatform/guest-test-infra/pull/307

hopkiw (approved):

michelle192837 (approved):

PR #1325, opened 2021-11-22
Add an alert for KES errors.

cjwagner:
/assign @listx @mpherman2


cjwagner:
/hold
So that I can apply this shortly after it merges.

PR #1393, opened 2022-01-06
pipelines: enable e2e test for frontend changes

zijianjoy:
So we can catch issues like this earlier https://oss-prow.knative.dev/view/gs/oss-prow/pr-logs/pull/kubeflow_pipelines/7137/kubeflow-pipeline-e2e-test/1478865285912989696

cc @chensun @Bobgy


chensun (commented):
/lgtm

Thanks!

PR #1398, opened 2022-01-08
pipelines: Merge two commands to one line for mkp-snapshot-test

zijianjoy:
The test doesn't generate any log, try to combine two commands to test it out.

https://pantheon.corp.google.com/storage/browser/_details/oss-prow/pr-logs/pull/kubeflow_pipelines/7153/kubeflow-pipeline-mkp-snapshot-test/1479705100409638912/build-log.txt;tab=live_object

@Bobgy


Bobgy:
Yes, they should be on the same line instead, because `-c` expects an entire string with the full script content.

/lgtm
/approve

PR #1399, opened 2022-01-10
pipelines: Use bash alpine for mktplace snapshot presubmit

zijianjoy:
Due to the snapshot test currently produces no log, I am switching to use bash:alpine image which allows me to run bash directly. It has the tool we need during the presubmit script, and matches the shebang in the shell script.
I also merged two commands into one line so the commands are wrapped in one quoted sentence following `-c` flag.

cc @chensun @Bobgy

### DIFF
--- a/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
+++ b/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
@@ -50,13 +50,12 @@ presubmits:
     decorate: true
     spec:
       containers:
-      - image: alpine:3.15.0
+      - image: bash:alpine3.14
         command:
-        - /bin/sh 
+        - /usr/local/bin/bash
         args:
         - -c
-        - cd ./manifests/gcp_marketplace/test
-        - sh ./presubmit.sh
+        - cd ./manifests/gcp_marketplace/test && bash ./presubmit.sh
   - name: kubeflow-pipeline-upgrade-test
     run_if_changed: "^(backend/.*)|(manifests/kustomize/.*)|(test/upgrade.*)$"
     cluster: build-kubeflow


prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
```diff
@@ -50,13 +50,12 @@ presubmits:
...
-      - image: alpine:3.15.0
+      - image: bash:alpine3.14
         command:
-        - /bin/sh 
+        - /usr/local/bin/bash
         args:
         - -c
-        - cd ./manifests/gcp_marketplace/test
-        - sh ./presubmit.sh
+        - cd ./manifests/gcp_marketplace/test && bash ./presubmit.sh
```
chensun, line 58: nit: I think if the image has bash, then the entire cmd can be simplified as `./manifests/gcp_marketplace/test/presubmit.sh`, just like what we have for the rest tests, right?

zijianjoy: If it can be simplified then it is awesome, but the presubmit test calls the snapshot.sh in the same file, where it is taking the `${DIR}` parameter, if we call presubmit from root, the `${DIR}` value is root, which caused the snapshot.sh not found: https://github.com/kubeflow/pipelines/blob/master/manifests/gcp_marketplace/test/presubmit.sh#L38

Bobgy: The DIR var is a common one-liner to find the folder containing the bash script, despite current working dir.
So you should be able to run it outside.

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
```diff
@@ -50,13 +50,12 @@ presubmits:
     decorate: true
     spec:
       containers:
-      - image: alpine:3.15.0
+      - image: bash:alpine3.14
         command:
-        - /bin/sh 
+        - /usr/local/bin/bash
```
Bobgy, line 55: it might be better omitting the full system path?
Just `bash` is enough.

Bobgy:
Another choice is installing bash right away in alpine image by

```
sh -c 'apk add bash && ./xxxx/yyyy.sh'
```

I'm not sure which is better, probably both solutions are similar.

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
zijianjoy: That is a good point, updated to use `bash` directly.

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
zijianjoy: I see, it is strange that ${DIR} doesn't work for me before, maybe it was because I was using /bin/sh instead of bash? Updated the command here.

zijianjoy:
> Another choice is installing bash right away in alpine image by
> 
> ```
> sh -c 'apk add bash && ./xxxx/yyyy.sh'
> ```
> 
> I'm not sure which is better, probably both solutions are similar.

That is a good point, I don't see difference between them functional-wise, but I will give it a try if the current PR doesn't work.

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
```diff
@@ -50,13 +50,12 @@ presubmits:
     decorate: true
     spec:
       containers:
-      - image: alpine:3.15.0
+      - image: bash:alpine3.14
         command:
-        - /bin/sh 
+        - bash
```
chensun, line 55: Synced offline, and James explained to me why we cannot have:
```yaml
command:
- ./script.sh
```
because in `bash:alpine3.14` image, the bash location is at `/usr/local/bin/bash` rather than `/bin/bash` which is what the script file has in its shebang.
In this case, I'd suggest either we switch to a different image that has the more common bash location so it saves the trouble and explanation, or we should add a comment here summarize what James found. Otherwise, the next person looking at this command might have the same confusion and go through the same investigation effort.

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
```diff
@@ -50,13 +50,12 @@ presubmits:
...
         args:
         - -c
-        - cd ./manifests/gcp_marketplace/test
-        - sh ./presubmit.sh
+        - bash ./manifests/gcp_marketplace/test/presubmit.sh
```
chensun, line 58: Could this be the following instead if we stick with the same image?
```yaml
command:
- bash
args:
- ./manifests/gcp_marketplace/test/presubmit.sh
```

`bash -c bash ./script.sh` seems redundant.

zijianjoy: That is a good point, removing the `-c` for removing redundant argument.

prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
zijianjoy: Makes sense, I haven't found an image which is small enough, and is stored at the desired location without explicitly downloading bash. I added command for our findings here. Does it look good to you?

chensun: Looking good. Thanks!

chensun (commented):
/lgtm
/approve

zijianjoy:
Thank you @chensun , would you like to LGTM/approve it again? I resolved the conflict with HEAD.

chensun:
/lgtm

Thanks!

PR #1400, opened 2022-01-10
Enable github workflows retest in knative repos

shinigambit:
Make use of the feature added here kubernetes/test-infra/issues/21581

### DIFF
--- a/prow/knative/plugins.yaml
+++ b/prow/knative/plugins.yaml
@@ -98,11 +98,8 @@ blunderbuss:
 triggers:
 - repos:
   - knative:
-    trigger_github_workflows: true
   - knative-sandbox:
-    trigger_github_workflows: true
-  - knative-prow-robot:
-    trigger_github_workflows: true
+  trigger_github_workflows: true
 
 # Plugins enabled for each repo.
 plugins:
--- a/prow/knative/plugins.yaml
+++ b/prow/knative/plugins.yaml
@@ -94,6 +94,16 @@ config_updater:
 blunderbuss:
   use_status_availability: true
 
+# Settings for the "trigger" plugin.
+triggers:
+- repos:
+  - knative:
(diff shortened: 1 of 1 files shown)


prow/knative/plugins.yaml
```diff
@@ -94,6 +94,16 @@ config_updater:
...
   use_status_availability: true
 
+# Settings for the "trigger" plugin.
+triggers:
+- repos:
+  - knative:
+    trigger_github_workflows: true
+  - knative-sandbox:
+    trigger_github_workflows: true
+  - knative-prow-robot:
```
chizhg, line 104: No need to add this since we used to add it for testing purpose. We can clean it up in follow-up PRs all together, though.

chizhg (commented):
🎉 🎉 🎉  Thanks @shinigambit !!!

/ok-to-test
/lgtm
/approve

prow/knative/plugins.yaml
```diff
@@ -94,6 +94,16 @@ config_updater:
 blunderbuss:
   use_status_availability: true
 
+# Settings for the "trigger" plugin.
+triggers:
+- repos:
```
chizhg, line 99: It looks `repos` should be a list of strings - https://github.com/kubernetes/test-infra/blob/e47f01cbed7a926b20f1dd4b7e18d06a6d9bc146/prow/plugins/config.go#L429

chizhg (changes requested):

prow/knative/plugins.yaml
```diff
@@ -94,6 +94,13 @@ config_updater:
...
+  - knative:
```
chaodaiG: `:` will be evaluated as part of the string matching so it won't work, dropping it would do

shinigambit:
Fixes knative/test-infra/issues/2992

PR #1422, opened 2022-01-24
gcp-blueprints: Run postsubmit only when it is master branch or release branch

zijianjoy:
Currently any PR on kubeflow/gcp-blueprints will trigger postsubmit test during PR creation. Therefore we should restrict to only the master branch and release branches.

### DIFF
--- a/prow/prowjobs/kubeflow/gcp-blueprints/kubeflow-gcp-blueprints.yaml
+++ b/prow/prowjobs/kubeflow/gcp-blueprints/kubeflow-gcp-blueprints.yaml
@@ -15,7 +15,7 @@ postsubmits:
   - name: kubeflow-gcp-blueprints-postsubmit
     cluster: build-kubeflow
     branches:
-    - ^master|.+-branch$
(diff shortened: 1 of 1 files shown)


prow/prowjobs/kubeflow/gcp-blueprints/kubeflow-gcp-blueprints.yaml
```diff
@@ -14,6 +14,8 @@ postsubmits:
   kubeflow/gcp-blueprints:
   - name: kubeflow-gcp-blueprints-postsubmit
     cluster: build-kubeflow
+    branches:
+    - ^master|.+-branch$
```
chensun, line 18: nit: maybe this can be a bit more strict on what release branch looks like? 
If I wasn't looking at the PR title at the same time I would wonder what `-branch` means. So probably worth a comment at least. WDYT?

zijianjoy: Make sense, I have added a comment to explain v1.1-branch format in gcp-blueprints repo.

PR #1428, opened 2022-01-25
test(pipelines): fix kubeflow-pipeline-backend-test

chensun:
golang:1.17.6-alpine3.15 doesn't have `bash` but `/bin/sh`

### DIFF
--- a/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
+++ b/prow/prowjobs/kubeflow/pipelines/kubeflow-pipelines-presubmits.yaml
@@ -20,6 +20,8 @@ presubmits:
       containers:
       - image: golang:1.17.6-alpine3.15
         command:
+        - /bin/sh
+        args:
         - ./test/presubmit-backend-test.sh
   - name: kubeflow-pipeline-e2e-test
     run_if_changed: "^(frontend/.*)|(backend/.*)|(proxy/.*)|(manifests/kustomize/.*)|(test/.*)$"


zijianjoy:
Can /bin/sh execute all the commands in https://github.com/kubeflow/pipelines/blob/master/test/presubmit-backend-test.sh?

chensun:
> Can /bin/sh execute all the commands in https://github.com/kubeflow/pipelines/blob/master/test/presubmit-backend-test.sh?

Yes, I think so. 
BTW the file will be changed by https://github.com/kubeflow/pipelines/pull/7210/files#diff-6d8f2d227947723178e123949208d2e65703f40d81032126b76ae38fa65e4338
The set of commands are the same though, including `wget`, `tar`, which I tested that both work under /bin/sh in the same image.

chensun:
> > Can /bin/sh execute all the commands in https://github.com/kubeflow/pipelines/blob/master/test/presubmit-backend-test.sh?
> 
> Yes, I think so. BTW the file will be changed by https://github.com/kubeflow/pipelines/pull/7210/files#diff-6d8f2d227947723178e123949208d2e65703f40d81032126b76ae38fa65e4338 The set of commands are the same though, including `wget`, `tar`, which I tested that both work under /bin/sh in the same image.

Actually, no. I missed there's a `git` command` not available in the image. Let me take another look see how to address it.

chensun:
> > > Can /bin/sh execute all the commands in https://github.com/kubeflow/pipelines/blob/master/test/presubmit-backend-test.sh?
> > 
> > 
> > Yes, I think so. BTW the file will be changed by https://github.com/kubeflow/pipelines/pull/7210/files#diff-6d8f2d227947723178e123949208d2e65703f40d81032126b76ae38fa65e4338 The set of commands are the same though, including `wget`, `tar`, which I tested that both work under /bin/sh in the same image.
> 
> Actually, no. I missed there's a `git` command` not available in the image. Let me take another look see how to address it.

Replaced `git diff` with `diff`, and verified it works under /bin/sh. Should be good now.
/unhold

zijianjoy:
/lgtm

Thank you Chen!

PR #1446, opened 2022-02-02
Fix Knative job config path.

cjwagner:
This PR (https://github.com/knative/test-infra/pull/3044) changed the Knative ProwJob config path which broke the config updating logic, the checkconfig job in this repo, and subsequently blocked [the Prow update](https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1439).

/assign @chizhg 
Please note that any config changes since the break have not been applied and a new change will be required to pick up the changes that were dropped.


chizhg:
Sorry for missing that, and thanks for the fix!

/lgtm

PR #1532, opened 2022-03-15
Update Knative max_prowjob_age from 24h to 240h

chizhg:
Knative has dot-release jobs that are scheduled every week, so with the current setting, most of the time the developers won't be able to find the jobs from prow.knative.dev and use Prow UI to retrigger them.

This PR updates Knative max_prowjob_age from 24h to 240h, which'll allow release managers to see more job history from Deck and retrigger the dot-release jobs. According to https://prow.knative.dev/ there are 1500+ jobs everyday, so x10 should not overload Prow since there were some recent performance improvements here.

I'm not aware of any other side effects, so please let us know if there are any especially those that can potentially increase the resource usage, since reducing the infra cost is our P0 task for CNCF infra migration initiatives. 

/cc @chaodaiG 

FYI @kvmware @carlisia @vsameer @dprotaso

### DIFF
--- a/prow/knative/config.yaml
+++ b/prow/knative/config.yaml
@@ -132,7 +132,7 @@ log_level: debug
 
 sinker:
   resync_period: 1m
-  max_prowjob_age: 24h
+  max_prowjob_age: 240h
   max_pod_age: 30m
   terminated_pod_ttl: 30m
 


chaodaiG:
I agree with the understanding of resource usage, and changing from 24 to 240h should not have much side effect. So:
/lgtm
/approve

However, there should be better ways of achieving rerun, the button is also available on spyglass pages, for example https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-serving-auto-release/1503707803942916096

Screen Shot 2022-03-15 at 12 49 20 PM
 
so
/hold

chizhg:
> I agree with the understanding of resource usage, and changing from 24 to 240h should not have much side effect. So: /lgtm /approve
> 
> However, there should be better ways of achieving rerun, the button is also available on spyglass pages, for example https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-serving-auto-release/1503707803942916096
> 
> Screen Shot 2022-03-15 at 12 49 20 PM
> 
> so /hold

The button is not shown on Spyglass pages for older Prow jobs, e.g. https://prow.knative.dev/view/gs/knative-prow/logs/ci-knative-serving-auto-release/1501291905713967104, is it a bug?

PR #1572, opened 2022-04-08
Add external secret from build cluster for compute-image-import

EricEdens:
Generated by running this script: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/master/prow/oss/create-build-cluster.sh


EricEdens:
Assigning to on-call based on CLI tool's instructions

/assign @fejta

PR #1581, opened 2022-04-18
Compute image import: Use main branch

EricEdens:
The job `compute-image-import-postsubmit-module-tests` currently isn't running: https://testgrid.k8s.io/googleoss-compute-image-import#module-tests . After debugging with pj-on-kind, I found that it was incorrectly configured: the repo `GoogleCloudPlatform/compute-image-import` uses the branch `main` instead of `master`.

Additionally, this removes a label which we aren't using for this cluster.


yswe (approved):
LGTM

MahmoudNada0:
LGTM

PR #1593, opened 2022-04-28
Fix 'grow' 'gprow' typo in ingress definition.

cjwagner:
Oops.
/assign @chaodaiG


chaodaiG (approved):

cjwagner:
This still isn't quite working. Apparently GKE ingresses only support global static IPs, but I created a regional one so I need to start over with a new static IP. I've begun that and reused the `oss-gprow-dev` static IP name so we shouldn't need to modify this again.

PR #1595, opened 2022-04-29
Adjust oauth config for new gprow.dev domains.

cjwagner:
When this merges I'll also complete the following:
- [ ] Update the oauth redirect settings in GitHub to use the new domain for the __private__ front end's oauth app.
- [ ] Update the oauth redirect settings in GitHub to use the new domain for the __public__ front end's oauth app.
- [ ] Update the oauth redirect URL in the oauth config K8s secret for the __private__ front end
- [ ] [later] Delete the old oauth config K8s secret for the __public__ front end. (Since we're using a more generally named secret.)

/assign @chaodaiG 
/hold

### DIFF
--- a/prow/oss/cluster/deck.yaml
+++ b/prow/oss/cluster/deck.yaml
@@ -31,7 +31,7 @@ spec:
         - --spyglass=true
         - --tide-url=http://tide/
         - --hook-url=http://hook:8888/plugin-help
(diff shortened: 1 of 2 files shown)


chaodaiG (approved):

fejta (approved):
/hold

PR #1618, opened 2022-05-13
Prow metrics collected with managed promethues

chaodaiG:
This is a new offering from GCP, which is managed prometheus services for collecting prometheus metrics. The configuration was pretty much copy/pasted from prow_podmonitors.yaml, with the following changes based on https://cloud.google.com/stackdriver/docs/managed-prometheus/setup-managed#gmp-pod-monitoring:

- Removed namespaceSelector section, which is not supported any more. The namespace is added on the PodMonitor CR itself.
- Renamed podMetricsEndpoints to endpoints.
- Use new apiVersion.


fejta (approved):
/hold

cjwagner (approved):

PR #1625, opened 2022-05-18
Break glass target for gencred refresher

chaodaiG:
This will help the case where gencred refresher failed more than 2 days.

/cc @cjwagner @listx @mpherman2

### DIFF
--- a/prow/oss/run-k8s-binary.sh
+++ b/prow/oss/run-k8s-binary.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# This script is for running arbitrary Go binaries from the k8s/test-infra repo.
+# Running this script when the binary cannot run inside a Docker container (for example,
+# when the binary needs to authenticate to GCP).
+# This script installs the binary directly from the k8s/test-infra repo instead of adding
+# k8s.io/test-infra as a Go module inside the current directory. This is because we've seen
+# dependency problems when doing so in other repos.
+#
+# Usage:
+# $1       : The relative path for installing Go binary from k8s/test-infra
+# $2..n    : The arguments for running the Go binary
+#
+# Example:
+# - ./run-k8s-binary.sh prow/cmd/generic-autobumper --config=prow/oss/oss-autobump-config.yaml
+
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}/../..")" && pwd -P)"
+
+TMP_DIR="$(mktemp -d)"
+K8S_INFRA_DIR="${TMP_DIR}/test-infra"
+BINARY="${TMP_DIR}/binary"
+
+# Install the Go binary directly from k8s/test-infra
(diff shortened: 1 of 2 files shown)


prow/oss/run-k8s-binary.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+# This script is for running arbitrary Go binary from k8s/test-infra repo,
+# running this script when this script cannot run docker container, for example
```
listx, line 3: Can you reword `running this script when this script cannot run docker container`? If I read this right it says that I should run this script when this script cannot run a docker container, but I don't think that's what you meant.

EDIT: Oh I think you meant to say `when the binary cannot run inside a Docker container`.

Here's my take at rewording this comment:

```
# This script is for running arbitrary Go binaries from the k8s/test-infra repo.
# Running this script when the binary cannot run inside a Docker container (for example,
# when the binary needs to authenticate to GCP).
# This script installs the binary directly from the k8s/test-infra repo instead of adding
# k8s.io/test-infra as a Go module inside the current directory. This is because we've seen
# dependency problems when doing so in other repos.
```

WDYT?

prow/oss/run-k8s-binary.sh
```diff
@@ -0,0 +1,34 @@
...
+
+TMP_DIR="$(mktemp -d)"
+K8S_INFRA_DIR="${TMP_DIR}/test-infra"
+BINARY="${TMP_DIR}/binary"
+
+# Install the Go binary directly from k8s/test-infra
+git clone git@github.com:kubernetes/test-infra.git "$K8S_INFRA_DIR"
+cd "$K8S_INFRA_DIR"
+go build -o "${BINARY}" "$1"
+shift 1
```
listx, line 31: Nit: `shift` alone is the same as `shift 1`.

listx (commented):
Wordsmithing and nit.

prow/oss/run-k8s-binary.sh
chaodaiG: looks nice, I'll steal it

prow/oss/run-k8s-binary.sh
chaodaiG: done

prow/oss/run-k8s-binary.sh
```diff
@@ -0,0 +1,34 @@
...
+BINARY="${TMP_DIR}/binary"
+
+# Install the Go binary directly from k8s/test-infra
+git clone git@github.com:kubernetes/test-infra.git "$K8S_INFRA_DIR"
+cd "$K8S_INFRA_DIR"
+go build -o "${BINARY}" "$1"
+shift
+
+cd $ROOT_DIR
+"$@"
```
cjwagner, line 34: I think we need to call the binary here, right? e.g.
`"${BINARY}" "$@"`

chaodaiG: good catch, fixed

cjwagner (approved):

PR #1643, opened 2022-06-09
Update_OneStepImageImport_e2e_tests_run_command

MahmoudNada0:
Update OneStepImageImport e2e tests run command:
- change the aws images IDs & path of cred file to WAW team aws account 
- remove aws-windows-ami
- add new aws-debian-ami

/cc yswe
/assign yswe

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/compute-image-import/compute-image-import.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/compute-image-import/compute-image-import.yaml
@@ -172,7 +172,7 @@ periodics:
       - -out_dir=$(ARTIFACTS)
       - -test_project_id=compute-import-test-pool-001
       - -test_zone=us-central1-b
-      - -variables=aws_region=us-east-2,aws_bucket=s3://onestep-test,ubuntu_ami_id=ami-04d75010218164863,windows_ami_id=ami-0c91f2e838828598d,ubuntu_vmdk=s3://onestep-test/ubuntu1804.vmdk,windows_vmdk=s3://onestep-test/windows2019.vmdk,aws_cred_file_path=gs://compute-image-import-test-resources/onestep-test-user,compute_service_account_without_default_service_account=pool-001-1-sa@compute-import-test-pool-001-1.iam.gserviceaccount.com,compute_service_account_without_default_service_account_permission=pool-001-2-sa@compute-import-test-pool-001-2.iam.gserviceaccount.com,project_id_without_default_service_account=compute-import-test-pool-001-1,project_id_without_default_service_account_permission=compute-import-test-pool-001-2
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleCloudPlatform/compute-image-import/compute-image-import.yaml
```diff
@@ -172,7 +172,7 @@ periodics:
       - -out_dir=$(ARTIFACTS)
       - -test_project_id=compute-import-test-pool-001
       - -test_zone=us-central1-b
-      - -variables=aws_region=us-east-2,aws_bucket=s3://onestep-test,ubuntu_ami_id=ami-04d75010218164863,windows_ami_id=ami-0c91f2e838828598d,ubuntu_vmdk=s3://onestep-test/ubuntu1804.vmdk,windows_vmdk=s3://onestep-test/windows2019.vmdk,aws_cred_file_path=gs://compute-image-import-test-resources/onestep-test-user,compute_service_account_without_default_service_account=pool-001-1-sa@compute-import-test-pool-001-1.iam.gserviceaccount.com,compute_service_account_without_default_service_account_permission=pool-001-2-sa@compute-import-test-pool-001-2.iam.gserviceaccount.com,project_id_without_default_service_account=compute-import-test-pool-001-1,project_id_without_default_service_account_permission=compute-import-test-pool-001-2
```
yswe, line 175: Do we need to pass here `debian_aim_id` as well, or do I miss anything?

prow/prowjobs/GoogleCloudPlatform/compute-image-import/compute-image-import.yaml
```diff
@@ -172,7 +172,7 @@ periodics:
...
+      - -variables=aws_region=us-west-2,aws_bucket=s3://onestep-e2e-tests,ubuntu_ami_id=ami-0f914731d0a56a859,debian_ami_id=ami-08b94fd8d55eea581,ubuntu_vmdk=s3://onestep-e2e-tests/aws-ubuntu-18.vmdk,windows_vmdk=s3://onestep-e2e-tests/windows-2019.vmdk,aws_cred_file_path=gs://compute-image-import-test-resources/onestep-e2e-cred-file,compute_service_account_without_default_service_account=pool-001-1-sa@compute-import-test-pool-001-1.iam.gserviceaccount.com,compute_service_account_without_default_service_account_permission=pool-001-2-sa@compute-import-test-pool-001-2.iam.gserviceaccount.com,project_id_without_default_service_account=compute-import-test-pool-001-1,project_id_without_default_service_account_permission=compute-import-test-pool-001-2
```
yswe, line 175: Do we need to pass `debian_ami_id` here as well, or do I miss anything?

yswe: Ah, my bad, you are actually passing it here!

We cannot remove `windows_ami_id` in this PR, because it is not a backward compatible change, so the tests will start failing. Could you keep `windows_ami_id` here and remove it in a new PR?

MahmoudNada0: I was planing to remove the windows-ami e2e test first and after it got merged we can merge this on the same time. So, this test will not fail. 
What do u think?

yswe: It will still start failing, because the `compute-image-import` project will wait for `debian_ami_id` which will not present until the changes are merged and rolled out in `oss-test-infra`.

The problem here is that we will start using a new AWS account where we won't have the `windows_ami_id` anyway, so as I see, we have to apply our changes in three steps:

1. Remove `windows_ami_id` test from `compute-image-import`;
2. Remove `windows_ami_id` from the variables in `oss-test-infra` and add `debian_ami_id`;
3. Add a new test with `debian_ami_id` to `compute-image-import`.

In this case we shouldn't face any failures.

MahmoudNada0: As agreed, we will make debian_ami_id optional for now to make sure that other tests will not affected if the run command in oss-test-infra doesn't has debian_ami_id flag.
But the Debian test will still fail until the changes in oss-test-infra are roll-out/released.

yswe (approved):

PR #1646, opened 2022-06-10
Remove private frontend for Looker

chaodaiG:
There is nothing left running on it now

cc @2nd-half


prow/oss/cluster/cluster.yaml
```diff
@@ -90,16 +90,6 @@ spec:
             name: hook
             port:
               number: 8888
-  - host: oss-prow-private.knative.dev
```
cjwagner, line 93: We can also remove the managed certs and associated ingress annotations too.

cjwagner (approved):
/hold

prow/oss/cluster/cluster.yaml
chaodaiG: good point, done, both

cjwagner (approved):

chaodaiG:
/unhold
other than updating ingress to not include private cert, this PR itself is pretty much noop.

will need to manually clean up the resources removed from this PR

PR #1683, opened 2022-07-11
Add prowjobs for kpt-config-sync

sdowell:


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
```diff
@@ -0,0 +1,149 @@
+presubmits:
+  GoogleContainerTools/kpt-config-sync:
+  - name: kpt-config-sync-presubmit
+    cluster: build-kpt-config-sync
+#    branches: ["master"]
```
janetkuo, line 5: Is this intentional?

sdowell: This is an artifact copied over from the reference job config, where it is also commented out

nan-yu: The branch filter was commented out so that the presubmit job can be triggered on any branches.

testgrid/config.yaml
```diff
@@ -11,6 +11,7 @@ dashboards:
 - name: googleoss-kubeflow-gcp-blueprints
 - name: googleoss-grpc-transcoder
 - name: googleoss-http-pattern-matcher
+- name: googleoss-kpt-config-sync
```
nan-yu, line 15: Does the dashboard group include all the prow jobs? I'm wondering how is it configured.

sdowell: To be honest I do not know all of the details about how this is configured. I put this in here to satisfy a presubmit check which was failing because there was a periodic job without an assigned dashboard.
```
* configuration error for (TestGroup) prober-runner-sa-key-rotate: Each Test Group must be referenced by at least 1 Dashboard Tab.
```

The presubmit jobs did not need to be annotated with a dashboard, so I left them unchanged. After doing some searching around it appears that the testgrid can be found [here](https://testgrid.k8s.io/googleoss). I think if we want the presubmit jobs to also show up on this dashboard that we can add them, but currently they do not have the dashboard annotation.

nan-yu: It is okay to not include the presubmit jobs in the dashboard as the result should show up in each PR. 
Just for my curiosity, are we configuring the postsubmit jobs in the same testgrid?

sdowell: We weren't planning to include any postsubmit jobs as of now, but if we did add postsubmit jobs I expect we would configure them in the same testgrid.

nan-yu: ACK.

chaodaiG (approved):

PR #1685, opened 2022-07-12
add kubeflow pipelines sdk code style checks

connor-mccarthy:
Adds kubeflow pipelines sdk code style checks, including yapf, docformatter, and isort. These checks are initially set as optional=true and skip_report=true, then will be both be set to false (default) in a future PR once code is proven to be adherent and tests are consistently passing.


chensun:
/lgtm
/approve

Thanks!

PR #1690, opened 2022-07-14
Add in-memory etcd label for docker in docker

sdowell:
Add docker-in-docker label that uses in-memory etcd to limit write
operations. This is currently intended for use with the config sync prow
jobs.


prow/oss/config.yaml
```diff
@@ -287,3 +287,15 @@ presets:
   volumeMounts:
   - name: docker-graph
     mountPath: /docker-graph
+- labels:
```
chaodaiG, line 290: Could you also add the PR description as comment above this line?

sdowell: done

chaodaiG (approved):

PR #1728, opened 2022-08-20
increase CS e2e memory request to 80Gi

sdowell:
The large-job-pool on our prow build cluster uses c2-standard-30 nodes,
which come with 30 vCPUs and 120GB memory. These jobs request 29 vCPUs
and 40Gi memory, meaning each job can only run on a single node but
underutilizes the memory on that node. This increases the memory request
to improve memory utilization on the node. This will hopefully reduce
test flakiness.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
@@ -71,8 +71,8 @@ presubmits:
           readOnly: true
         resources:
           requests:
-            # When running at parallel=15, we use 20Gi. So request 40Gi to be safe.
-            memory: "40Gi"
+            # When running at parallel=15, we use 20Gi. So request 80Gi to be safe.
+            memory: "80Gi"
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
```diff
@@ -128,8 +128,8 @@ presubmits:
           readOnly: true
         resources:
           requests:
-            # When running at parallel=15, we use 20Gi. So request 40Gi to be safe.
-            memory: "40Gi"
+            # When running at parallel=15, we use 20Gi. So request 80Gi to be safe.
```
karlkfi, line 131: "we use 20 so request 80" sounds funny. Do you know how much it uses now?

sdowell: I haven't measured how much it uses now, so I can't say for sure. At any rate our nodes are underutilized so it should be a safe change

karlkfi (approved):
/lgtm

PR #1735, opened 2022-08-25
skip e2e on non-functional changes

sdowell:


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
```diff
@@ -28,7 +28,8 @@ presubmits:
   - name: kpt-config-sync-presubmit-e2e-mono-repo
     cluster: build-kpt-config-sync
     # branches: ["master"]
-    always_run: true
+    always_run: false
+    skip_if_only_changed: "\\.md$|^(LICENSE|OWNERS)$"
```
nan-yu, line 32: Can we add the `docs` and `dashboard` folders to the skip list?

sdowell: PR was automerged, but yeah we can add those as a followup

PR #1740, opened 2022-09-08
increase CS e2e max_concurrency to 5

sdowell:
The node pool can support running up to 10 concurrent jobs with the current resource requests. This bumps the max concurrency to improve cluster utilization and increase how many presubmits can run concurrently.


nan-yu (approved):
/lgtm

PR #1743, opened 2022-09-12
Increase the memory request for e2e submit job

haiyanmeng:
While testing the parallel setting of the CS presubmit mono/multi jobs, the test pod got deleted due to memory pressure. 
This node used for the presubmit jobs has 112GiB memory. This PR increases the memory request from 80GiB to 105GiB to see whether it helps relieve the memory pressure.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
@@ -71,8 +71,7 @@ presubmits:
           readOnly: true
         resources:
           requests:
-            # When running at parallel=15, we use 20Gi. So request 80Gi to be safe.
-            memory: "80Gi"
+            memory: "105Gi"
             # This value is experimentally determined.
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-presubmits.yaml
```diff
@@ -71,8 +71,7 @@ presubmits:
           readOnly: true
         resources:
           requests:
-            # When running at parallel=15, we use 20Gi. So request 80Gi to be safe.
-            memory: "80Gi"
+            memory: "108Gi"
```
sdowell, line 74: I think `108Gi` might not leave enough memory for other system pods to run on the node. Only `109.12Gi` is allocatable on the node, so this leaves just ~1 GB for other pods. Maybe we can try something around 100-105 GB?

haiyanmeng: Sounds good. Updated the PR.

sdowell (approved):

PR #1750, opened 2022-09-19
Adjust the CPU/mem resources

haiyanmeng:
The peak CPU usage of the presubmit mono/multi jobs is about 10 CPUs; The peak mem usage of the presubmit mono/multi jobs is about 16 GiBs.

This PR sets the CPU/mem requests to 17CPUs/30GiBs, so that we can support running three presubmit jobs on a single node concurrently.


sdowell:
Are you confident this won't cause stability issues?

sdowell (approved):
/lgtm

PR #1755, opened 2022-09-20
Increase max_concurrency of mono/multi presubmit jobs from 3 to 4

haiyanmeng:
Currently, a PR triggers 3 mono presubmit jobs, and 3 multi presubmit jobs. Every mono/multi presubmit job requests about 50% of the CPU/mem resources of a node. So running mono/multi presubmit jobs for a PR requires 3 nodes. Our node pool currently has 12 nodes, and can support running presubmit jobs for 4 PRs at the same time.


nan-yu (approved):
/lgtm

PR #1758, opened 2022-09-21
use yaml anchors for config sync e2e tests

sdowell:


nan-yu (approved):
/lgtm

Thanks for the update. @rquitales FYI, we're expecting similar configs for periodic jobs.

PR #1765, opened 2022-09-26
fix commandHelpLink for approve plugin

sdowell:
The approve plugin by default links to go.k8s.io/bot-commands, which takes the user to https://prow.k8s.io/command-help rather then the oss prow deck. This updates the plugin configuration to direct the user to the command help page for oss.gprow.dev.

### DIFF
--- a/prow/oss/plugins.yaml
+++ b/prow/oss/plugins.yaml
@@ -34,6 +34,7 @@ approve:
   - grpc-ecosystem
   - chaotoppicks
   require_self_approval: false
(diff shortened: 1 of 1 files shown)


sdowell:
This change should impact all repos
/assign @cjwagner

cjwagner (approved):
FYI this page and the plugin help page can be slow to load on this instance due to the large number of repos in the `google` and `GoogleCloudPlatform` GitHub organizations. If the page fails to load try refreshing after a few seconds once the repo information has a minute to be fetched and cached.

PR #1766, opened 2022-09-26
remove prober-cred volume mount for presubmit jobs

sdowell:
This service account is not used for presubmit jobs since the presubmits run in kind. The service account is for the periodic jobs which target GKE clusters.


nan-yu (approved):
/lgtm

PR #1772, opened 2022-09-28
Configure environment variables for the kpt-config-sync CI

nan-yu:


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -21,12 +21,17 @@ prow_ignored:
       command:
       - runner.sh
       env:
+      # This is used to tag and push the Config Sync images to GCR.
+      - name: USER
```
sdowell, line 25: I was thinking about a more intuitive way we can track and isolate docker tags we use in CI. I was thinking we could use something like `BUILD_ID` or `PROW_JOB_ID` and default to previous `USER` behavior (for local dev usage) if neither of those variables are set. I think `BUILD_ID` would make sense. This would require some changes to our make target functionality
https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables

WDYT?

nan-yu: SG. DONE. Removed USER

PR #1775, opened 2022-09-29
Make the dummy test jobs always green

nan-yu:
We're going to replace these two dummy test jobs with real CI jobs. Noop them in the interim.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -2,7 +2,8 @@
 # to view expanded yaml, run: `yq eval 'explode(.)' [file]`
 prow_ignored:
 - &config-sync-ci-job
-  interval: 1h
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -58,9 +58,9 @@ periodics:
     containers:
     - <<: *config-sync-ci-container
       args:
-      - make
-      - test-e2e-gke-multi-repo-test-group1
-      - 'GCP_CLUSTER=multi-repo-gke-test-group-1-std-stable'
+      #  We're going to replace these two dummy test jobs with real CI jobs.
+      #  Noop them in the interim.
+      - exit 0
```
sdowell, line 63: I think you might have to put the exit code as a separate arg

nan-yu: DONE. I also reduced the interval from 1h to 1d.

PR #1780, opened 2022-10-04
set GCP_PROJECT to stolos-dev in presubmits

sdowell:
the prow project has not yet been set up with all of the resources needed to run the tests


nan-yu (approved):
/lgtm

PR #1786, opened 2022-10-05
experimental: use workload identity for periodic

sdowell:
This change updates the job spec for one job to use workload identity. Only one job is changed so this can be tested without impacting the entire test grid.


nan-yu (commented):
The change alone won't make the test pass as the make target requires a mounted SA key file. We need to update the make target to not reply on the SA key file first.

nan-yu (approved):
/lgtm

PR #1787, opened 2022-10-05
use empty string for PROBER_DOCKER_ARGS

sdowell:
the quotes were being expanded to the docker call, causing the execution to fail with invalid reference format error.


nan-yu (approved):
/lgtm

PR #1790, opened 2022-10-06
switch periodic jobs to use workload identity

sdowell:
Tested workload identity on one job and it passed, so pushing the config to the rest of the jobs. If this is stable then the job that rotates the secret can be removed in a subsequent change.


nan-yu (approved):
/lgtm

PR #1793, opened 2022-10-11
Update ESPv2 kubekins images

nareddyt:
Signed-off-by: nareddyt <nareddyt@google.com>


nareddyt:
> Can we find any doc describing these images?

Unfortunately I couldn't find any docs. I have to ask around internally to even understand where the images were moved to.

They are built in this repo: https://github.com/kubernetes/test-infra

PR #1795, opened 2022-10-12
Add more CI tests for config-sync

nan-yu:


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -1012,3 +1012,74 @@ periodics:
...
+  annotations:
+    testgrid-dashboards: googleoss-kpt-config-sync-misc
+    testgrid-tab-name: multi-repo-kind
+  spec:
+    <<: *config-sync-ci-job-spec
+    containers:
+    - <<: *config-sync-ci-container
+      args:
+      - make
+      - test-e2e-kind-multi-repo
```
sdowell, line 1027: It occurred to me that these tests run directly on the prow build cluster in the large job pool rather than on a separate GKE cluster. I think this has a few implications:
- We should remove the kind gke clusters from our [terraform config](https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/e2e/testinfra/terraform/prow/prow_clusters.tf#L110)
- We need to account for these periodic jobs in our large job pool. We currently set the max_concurrency for the presubmit jobs based on the available resources in the large job pool, and this will impact that calculation.

I have some ideas about how we can make the large job pool more manageable, which perhaps we can discuss offline. I think the simplest workaround is that we could create a new node pool and use a separate node selector for these periodics, although I think there may be better options.

nan-yu: Yes, we need to remove the kind GKE clusters.
And you're right. I forgot we need to use the large job pool. Can we make the kind jobs run on a cluster that is separate from the presubmit cluster? The separation will avoid recalculating the resources.

sdowell: I think we can just use a separate node pool with a separate node pool selector - that would be the simplest solution that doesn't deviate much from the current approach. We took a similar approach for the `v1.13` release branch presubmit jobs. I don't think there would be a straightforward way to run them in a separate cluster from the prow build cluster

nan-yu: Had a discussion with Sam offline. We decided to go with the separate node pool for the kind jobs. Let me know when the new node pool is ready, and then I can update the prow configs accordingly.

nan-yu: I removed the kind jobs from the current PR and will have a new PR for them.

PR #1796, opened 2022-10-12
Add prow jobs for config-sync mono-repo

nan-yu:


nan-yu:
We are going to freeze the mono-repo codebase and move it to GoB, which implies that we're going to move these jobs back to GoB. Closing this issue for now.

PR #1806, opened 2022-10-17
use custom labels for presubmit nodeSelector

sdowell:
Using the node pool label makes it difficult to recreate the node pool or incorporate more machine types into the node pool. This change uses a custom label selector to enable using multiple node pools for the kind presubmit jobs.


nan-yu (approved):
/lgtm

Issue #1818, opened 2022-10-27
post-oss-test-infra-deploy-prow job regularly failing


chases2:
Jobs are failing with the following error:
```
note: module requires Go 1.17
make[2]: *** [Makefile:63: install] Error 2
make[2]: Leaving directory '/home/prow/go/src/github.com/GoogleCloudPlatform/oss-test-infra/prow/oss/cluster/monitoring/mixins'
make[1]: *** [Makefile:26: apply] Error 2
make[1]: Leaving directory '/home/prow/go/src/github.com/GoogleCloudPlatform/oss-test-infra/prow/oss/cluster/monitoring'
make: *** [Makefile:46: deploy-monitoring] Error 2
make: Leaving directory '/home/prow/go/src/github.com/GoogleCloudPlatform/oss-test-infra/prow/oss'
```
Excerpt from https://oss.gprow.dev/view/gs/oss-prow/logs/post-oss-test-infra-deploy-prow/1585738959382646784

Prowjob Definition: https://github.com/GoogleCloudPlatform/oss-test-infra/tree/master/prow/prowjobs//GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml
Job History: https://testgrid.k8s.io/googleoss-test-infra#prow-deploy

Updating the job to the latest version didn't work; the underlying image may need to be upgraded: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/1817

chases2:
Have confirmed that the make command the image is running works locally. Trying to figure out what's different between the two environments.

chases2:
Ah, the image source was bumped to v1.17 in 2022: https://github.com/kubernetes/test-infra/commit/023af195417224c7e34c50d1f695952bc8584573

But the most recent image was built in 2021, so it probably contains an earlier version: gcr.io/k8s-testimages/gcloud-in-go:v20210913-fc7c4e8

chases2:
Yayy https://oss.gprow.dev/view/gs/oss-prow/logs/post-oss-test-infra-deploy-prow/1586081077275922432

PR #1819, opened 2022-10-28
Use k8s-staging-test-infra version of gcloud-in-go

chases2:
The old repository's image is no longer being updated.

issue https://github.com/GoogleCloudPlatform/oss-test-infra/issues/1818


chases2:
Note: it doesn't appear that we're using gcloud-in-go image anywhere else in this repository.

`grep -r gcloud-in-go`

chaodaiG (approved):

PR #1826, opened 2022-11-07
remove monorepo presubmits

sdowell:
With the monorepo code migrated to a separate repo, we no longer need to run the monorepo presubmits on kpt-config-sync. Note that with the reduction in prowjobs per presubmit we require fewer resources in a presubmit run. We could potentially increase the job concurrency accordingly, but for now I intend to scale down the node pool instead. We can scale the node pool back up later if we desire a higher job concurrency.


xinnywinne (approved):
/lgtm

PR #1827, opened 2022-11-07
remove monorepo periodics

sdowell:
With the monorepo code migrated to a separate repo, we no longer need to run the monorepo periodics on kpt-config-sync.


xinnywinne (approved):
/lgtm

PR #1831, opened 2022-11-09
Update jwt_verify_lib CI image

orionHong:
Image is created from https://github.com/google/jwt_verify_lib/pull/89.

### DIFF
--- a/prow/prowjobs/google/jwt_verify_lib/jwt-verify-lib-presubmit.yaml
+++ b/prow/prowjobs/google/jwt_verify_lib/jwt-verify-lib-presubmit.yaml
@@ -10,7 +10,7 @@ presubmits:
       description: "Runs all unit tests per PR."
     spec:
       containers:
-      - image: gcr.io/cloudesf-testing/gcpproxy-prow:v20220829-v2.38.0-5-gc746373d-master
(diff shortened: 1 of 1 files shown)


prow/prowjobs/google/jwt_verify_lib/jwt-verify-lib-presubmit.yaml
```diff
@@ -10,7 +10,7 @@ presubmits:
       description: "Runs all unit tests per PR."
     spec:
       containers:
-      - image: gcr.io/cloudesf-testing/gcpproxy-prow:v20220829-v2.38.0-5-gc746373d-master
+      - image: gcr.io/cloudesf-testing/gcpproxy-prow:v20221109-jwt-verify-lib
```
qiwzhang, line 13: instead of tag it as `jwt_verify_lib`,   maybe we should call it out in the name,  e.g. `gcr.io/cloudesf-testing/jwt-verify-lib-prow`

orionHong: That's a good idea. Done :)

qiwzhang (approved):

PR #1842, opened 2022-11-23
chore: Increase timeout for kpt-config-sync autopilot periodics

rquitales:
The current periodics can take up to 4 hours to compelete which gets interrupted by the current 4 hour timeout. This change increases the timeout to reduce test flakiness.


nan-yu (approved):
/lgtm

PR #1846, opened 2022-11-28
[ESPv2] Update prow env to image with LLVM 14

nareddyt:
Latest envoy only builds with clang-14

Dockerfile changes in https://github.com/GoogleCloudPlatform/esp-v2/pull/755


nareddyt:
@TAOXUY @orionHong or @qiwzhang PTAL

qiwzhang (approved):

PR #1851, opened 2022-12-07
Adding PubSub subscriptions and prow jobs for Louhi

k8kim:
This adds
    - PubSub subscriptions to louhi-resources/oss-prow-trigger 
    - Periodic prow job for Louhi that just echoes the date of the day

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/louhi/louhi-echo-test.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/louhi/louhi-echo-test.yaml
@@ -2,6 +2,8 @@ periodics:
 - cron: "* * 30 02 *" # Invalid date, should only run when triggered
   name: louhi-echo-test-periodic
   decorate: true
+  annotations:
+    testgrid-create-test-group: "false"
   spec:
     containers:
     - image: alpine


prow/oss/config.yaml
```diff
@@ -264,6 +264,9 @@ gerrit:
   - org: https://linux-review.googlesource.com
     repos:
     - linux/kernel/git/torvalds/linux
+  - org: https://cloud-release-internal-review.googlesource.com
```
chaodaiG, line 267: I'm not sure that this prow instance has access to this Gerrit host. Since this PR is just for testing pubsub a GitHub repo should be sufficient, would suggest remove this entry and I'll take a look at which GitHub repo you can use for this one time test

prow/prowjobs/GoogleCloudPlatform/louhi/louhi-echo-test.yaml
```diff
@@ -0,0 +1,19 @@
+periodics:
+- cron: "* * 30 02 *" # Invalid date, should only run when triggered
+  name: louhi-echo-test-periodic
+  decorate: true
+  spec:
+    containers:
+    - image: alpine
+      command: ["/bin/date"]
+postsubmits:
+  https://cloud-release-internal-review.googlesource.com/louhi/test-prow:
```
chaodaiG, line 10: let's try it out in my personal dev repo, replacing this with `chaotoppicks/great-tools`

prow/prowjobs/GoogleCloudPlatform/louhi/louhi-echo-test.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - name: louhi-echo-test-postsubmit
+    cluster: test-infra-trusted
```
chaodaiG, line 12: let's remove this, use default build cluster instead

prow/oss/config.yaml
k8kim: I will actually get rid of gerrit stuff, and just use periodic job.

prow/prowjobs/GoogleCloudPlatform/louhi/louhi-echo-test.yaml
k8kim: Will get rid of postsubmit jobs

prow/prowjobs/GoogleCloudPlatform/louhi/louhi-echo-test.yaml
k8kim: 👍

PR #1863, opened 2023-01-04
add Config Sync release periodics

sdowell:
This is a similar set of periodic jobs that we run on the main branch. This set of periodics is intended to verify the current/latest release branch.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
@@ -1,0 +1,1080 @@
+# YAML anchors for reference elsewhere.
+# to view expanded yaml, run: `yq eval 'explode(.)' [file]`
+prow_ignored:
+- &config-sync-ci-job
+  interval: 1h
+  cluster: build-kpt-config-sync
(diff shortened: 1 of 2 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
```diff
@@ -0,0 +1,1320 @@
...
+    testgrid-tab-name: standard-regular-gitlab
+  spec:
+    <<: *config-sync-ci-job-spec
+    containers:
+    - <<: *config-sync-ci-container
+      args:
+      - make
+      - test-e2e-gke-multi-repo-test-group1
+      - 'E2E_ARGS=--git-provider=gitlab'
+      - 'GCP_CLUSTER=multi-repo-1-standard-regular-gitlab'
```
nan-yu, line 194: We can skip the bitbucket and gitlab tests for the release CI because they have the same coverage for CS functionalities.

sdowell: Good catch - it wasn't part of our plan to include these. Removed

nan-yu (approved):
/lgtm

PR #1864, opened 2023-01-05
Update kubeflow-pipelines.presubmits.yaml to use a node14 and not node12

ryansteakley:
Node12 is long deprecated and should move to use node14


zijianjoy:
/lgtm
/approve

Thank you @ryansteakley for the change!

PR #1910, opened 2023-03-10
ci: bump kubekins-e2e image

sdowell:
This uses an upgraded kubekins-e2e base image which contains Go 1.19.

Uses the image defined in https://github.com/GoogleContainerTools/kpt-config-sync/pull/466


nan-yu (approved):
/lgtm

PR #1923, opened 2023-03-30
remote v113 presubmit jobs

sdowell:
Cherry picks are done for v113, so these job definitions can be removed. Removing these will enable us to spin down the legacy node pool.


nan-yu (approved):
/lgtm

PR #1955, opened 2023-05-19
increase frequency of vuln scan job

sdowell:
This job is relatively lightweight, so we can afford to run it more frequently. This will enable faster feedback on changes/fixes.


karlkfi (approved):
/lgtm

PR #1967, opened 2023-06-12
add config sync periodics for parallel gke

sdowell:
These targets use the newly added entrypoints for running tests in parallel on GKE. For now these jobs are experimental to discover any issues with the configuration, but the intention is to replace the testgroup-based jobs with this new pattern.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -1343,6 +1343,101 @@ periodics:
       - 'E2E_ARGS=--stress -run=TestStress*'
       - 'GCP_CLUSTER=stress-test'
 
+# Experimental prowjob to verify the new create-clusters flow on standard.
+# If stable, this is intended to replace the above testgroup-based definitions
+- name: standard-regular
+  interval: 1h
+  annotations:
+    testgrid-dashboards: googleoss-kpt-config-sync-misc
+    testgrid-tab-name: standard-regular
+  cluster: build-kpt-config-sync
+  decorate: true
+  decoration_config:
+    timeout: 2h
+  extra_refs:
+  - org: GoogleContainerTools
+    repo: kpt-config-sync
+    base_ref: main
+  spec:
+    serviceAccountName: e2e-test-runner
+    containers:
+    - image: us-docker.pkg.dev/kpt-config-sync-ci-artifacts/test-infra/gke-e2e:v1.0.0-d11cc175e
+      command:
+      - make
+      args:
+      - test-e2e-gke-ci
+      env:
+      - name: GKE_E2E_TIMEOUT
+        value: 2h
+      - name: GCP_PROJECT
+        value: kpt-config-sync-ci-main
+      - name: GCP_NETWORK
+        value: prow-e2e-network-1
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -1343,6 +1343,101 @@ periodics:
...
+      - name: GCP_NETWORK
+        value: prow-e2e-network-1
+      - name: GCP_SUBNETWORK
+        value: prow-e2e-subnetwork-1
+      - name: GCP_ZONE
+        value: us-central1-a
+      - name: GKE_RELEASE_CHANNEL
+        value: regular
+      - name: NUM_CLUSTERS
+        value: "8"
```
sdowell, line 1383: I'm planning to experiment with scaling these up, but for now I want to use the same level of parallelism as the existing CI so we can do an apples to apples comparison

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -1343,6 +1343,101 @@ periodics:
       - 'E2E_ARGS=--stress -run=TestStress*'
       - 'GCP_CLUSTER=stress-test'
 
+# Experimental prowjob to verify the new create-clusters flow on standard.
+# If stable, this is intended to replace the above testgroup-based definitions
+- name: standard-regular
+  interval: 1h
```
karlkfi, line 1349: do we want to run this every hour? how long does it usually take with 8 parallel? Should we start less frequent and ramp it up when we know?

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -1343,6 +1343,101 @@ periodics:
...
+  annotations:
+    testgrid-dashboards: googleoss-kpt-config-sync-misc
+    testgrid-tab-name: standard-regular
+  cluster: build-kpt-config-sync
+  decorate: true
+  decoration_config:
+    timeout: 2h
```
karlkfi, line 1356: will it start another one if the previous is still running?

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
sdowell: This is effectively the same parallelism into the existing tests which are sharded into 8 testgroups, so it theoretically should take approximately the average runtime of the current jobs. I haven't done extensive timing but the runtime is around an hour. The 1h interval in practice means the tests are always running

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
sdowell: My understanding is that periodic jobs won't run concurrently, i.e. it will wait for the previous to complete. The docs seem to confirm this
```
	// Consecutive jobs are run at `interval` duration apart, provided the
	// previous job has completed.
```

karlkfi (approved):
/lgtm

PR #1971, opened 2023-06-13
explicitly unset GCP_ZONE for autopilot

sdowell:
Only one of GCP_ZONE or GCP_REGION can be specified for creating a cluster. The makefile by default supplies a GCP_ZONE and must be unset.


karlkfi (approved):
/lgtm

PR #1979, opened 2023-06-28
update experimental e2e jobs to use lazy create

sdowell:
This updates the CI jobs to use create-clusters=lazy and destroy-clusters=false. This means the job will create the named clusters if they do not exist, and then leave them for the next execution to reuse.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -41,6 +41,50 @@ prow_ignored:
       # This job requires 8vCPUs or less, so it is "small".
       cloud.google.com/gke-nodepool: small-job-pool
 
+# Experimental anchor to verify the new create-clusters flow.
+# If stable, this is intended to replace the above anchor.
+- &config-sync-ci-job-v2
+  interval: 1h
+  cluster: build-kpt-config-sync
+  decorate: true
+  decoration_config:
+    timeout: 4h
+  extra_refs:
+  - org: GoogleContainerTools
+    repo: kpt-config-sync
+    base_ref: main
+  spec: &config-sync-ci-job-spec-v2
+    serviceAccountName: e2e-test-runner
+    containers:
+    - &config-sync-ci-container-v2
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -41,6 +41,48 @@ prow_ignored:
       # This job requires 8vCPUs or less, so it is "small".
       cloud.google.com/gke-nodepool: small-job-pool
 
+- &config-sync-ci-job-v2
```
karlkfi, line 44: maybe add a comment about how v2 is different?

sdowell: added a comment

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -1345,101 +1389,42 @@ periodics:
...
-      resources:
-        requests:
-          memory: "8Gi"
-          cpu: "4000m"
-    nodeSelector:
-      # This job requires 8vCPUs or less, so it is "small".
-      cloud.google.com/gke-nodepool: small-job-pool
+      - 'GKE_E2E_TIMEOUT=2h'
+      - 'GCP_SUBNETWORK=prow-e2e-subnetwork-1'
+      - 'GCP_ZONE=us-central1-a'
```
karlkfi, line 1404: we sure we want to use this zone?
is there another that might be less likely to have capacity outages and slow cluster creation?

sdowell: the zone we're using for standard isn't causing any major issues, it's the regional autopilot clusters that sometimes timeout.
See:
- autopilot: https://testgrid.k8s.io/googleoss-kpt-config-sync-misc#autopilot-regular
- standard: https://testgrid.k8s.io/googleoss-kpt-config-sync-misc#standard-regular

I expect using the lazily created clusters should resolve flakiness associated with cluster creation. I'm not familiar with data on which regions may have more capacity, I expect we would need to do some experimentation. However, even with a different region I expect there will always be times that we would experience capacity issues

karlkfi (approved):
/lgtm

PR #1981, opened 2023-06-29
increase interval for new cs e2e jobs

sdowell:
The new job types run faster than the old ones, the e2e tests on standard now can often complete in ~45 minutes. Decreasing the interval will improve the turnaround time on the new periodics.


karlkfi (approved):
/lgtm

PR #1986, opened 2023-07-06
add standard-rapid CS periodic using new job type

sdowell:
This expands the new job type to the rapid channel on standard GKE. This also experiments with a new cluster shape which uses faster compute/storage but fewer overall resources.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -1406,6 +1406,50 @@ periodics:
       - 'E2E_NUM_CLUSTERS=10'
       - 'E2E_CLUSTER_PREFIX=standard-regular'
 
+# Experiment with new cluster shape to use faster resources with lower footprint.
+# Uses a single node cluster to minimize number of persistent disks.
+- <<: *config-sync-ci-job-v2
+  name: kpt-config-sync-standard-rapid
+  annotations:
+    testgrid-dashboards: googleoss-kpt-config-sync-misc
+    testgrid-tab-name: standard-rapid
+  spec:
+    <<: *config-sync-ci-job-spec-v2
+    containers:
+    - <<: *config-sync-ci-container-v2
+      args:
+      - 'GKE_E2E_TIMEOUT=2h'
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -1406,6 +1406,29 @@ periodics:
...
+    - <<: *config-sync-ci-container-v2
+      args:
+      - 'GKE_E2E_TIMEOUT=2h'
+      - 'GCP_SUBNETWORK=prow-e2e-subnetwork-3'
+      - 'GCP_ZONE=us-central1-a'
+      - 'GKE_RELEASE_CHANNEL=rapid'
+      - 'E2E_NUM_CLUSTERS=10'
+      - 'E2E_CLUSTER_PREFIX=standard-rapid'
+      - 'GKE_NUM_NODES=1'
+      - 'GKE_MACHINE_TYPE=n2-standard-4'
```
karlkfi, line 1428: do the gke daemons, addons, and e2e tests all fit on a 4 core machine? I would be really surprised...

sdowell: Yeah I ran with this config locally and it succeeded. I can choose a bigger machine, although our current 12 vCPU / 48 Gb clusters are pretty overkill/underutilized

sdowell: we use e2-standard-4 x 3, so 3 nodes with 4 cores each

sdowell: I have an idle cluster (no tests running) with this configuration and there is 1.5 vCPU requested, which leaves ~2.5 vCPU for the Config Sync workloads to run on. I think n2-standard-16 would be pretty overkill, but maybe n2-standard-8 would be a middle ground that leaves some headroom.

I ran the whole test suite with this config and it passed, so it should be sufficient. But I haven't ran it extensively to test how stable it is, hence why I'm experimenting with the prowjob

karlkfi (approved):
/lgtm

PR #1988, opened 2023-07-07
move new CS periodics to main dashboard

sdowell:
This is the intended dashboard format, where the periodics against the main branch will be consolidated into a single dashboard. The legacy testgroup based prowjobs and associated dashboards will be cleaned up once all of the new jobs are set up.


karlkfi (approved):
/lgtm

PR #1991, opened 2023-07-10
set cluster shape for new jobs

sdowell:
Experimentally this shape is stable and the fastest of the configurations that were tested. It also consumes a lower overall resource footprint compared to the other configurations that were tested.


karlkfi (approved):
/lgtm

PR #1992, opened 2023-07-10
add remaining CS prowjobs using new job type

sdowell:
This implements all existing periodics using the new prow job type. Once these are all running and stable, we will begin retiring the old jobs.


sdowell:
/hold
Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/730 for Config Controller cluster support.
/assign @karlkfi

karlkfi (approved):
/lgtm

PR #1994, opened 2023-07-11
Change iso to windows 2019 core

koln67:
Fix windows iso for presubmit


dorileo:
As a side note: I'm approving this PR so we can unblock pipeline changes testing. I'd like to see a followup PR migrating this definitions to use a gcloud secret.

PR #1999, opened 2023-07-14
chore: bump CS kubekins-e2e image

sdowell:
The image has been redefined for the new test-infra registry, but is essentially the same image.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
@@ -20,7 +20,7 @@ prow_ignored:
     serviceAccountName: e2e-test-runner
     containers:
     - &config-sync-ci-container
-      image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20230309-9a6b1b3121-1.23
+      image: us-docker.pkg.dev/kpt-config-sync-ci-artifacts/test-infra/kubekins-e2e:v1.0.0-612a3a80d
       command:
       - runner.sh
       env:
@@ -1048,7 +1048,7 @@ periodics:
     <<: *config-sync-ci-job-spec
     containers:
     - <<: *config-sync-ci-container
-      image: gcr.io/oss-prow-build-kpt-config-sync/e2e-prow:kubekins-e2e-v20230309-9a6b1b3121-1.23-kind-v0.14.0
+      image: us-docker.pkg.dev/kpt-config-sync-ci-artifacts/test-infra/kubekins-e2e:v1.0.0-612a3a80d
       args:
       - make
       - test-e2e-kind-multi-repo
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -19,7 +19,7 @@ prow_ignored:
     serviceAccountName: e2e-test-runner
     containers:
     - &config-sync-ci-container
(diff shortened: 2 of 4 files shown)


haiyanmeng (approved):

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
```diff
@@ -20,7 +20,7 @@ prow_ignored:
     serviceAccountName: e2e-test-runner
     containers:
     - &config-sync-ci-container
-      image: gcr.io/k8s-staging-test-infra/kubekins-e2e:v20230309-9a6b1b3121-1.23
+      image: us-docker.pkg.dev/kpt-config-sync-ci-artifacts/test-infra/kubekins-e2e:v1.0.0-612a3a80d
```
nan-yu, line 23: Is this a tag and repush?

sdowell: Use this make target: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/Makefile.oss.prow#L37

nan-yu: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/build/prow/kubekins-e2e/Dockerfile is just the base image, so it is basically a tag and repush. Why don't we use the k8s-staging-test-infra one directly?

sdowell: Having the tooling in place makes it straightforward to make modifications to the image should we need to (e.g. this sequence of changes consolidated our `kubekins-with-kind` variant into a single image). I think it's also good practice to host the images in a registry we control to avoid unexpected breakage/outage from upstream. If you feel strongly though we can use the image from the k8s registry directly

nan-yu: I'm not strongly against the process, just trying to understand why it is different from other repositories. For resource-group, we use the k8s-staging-test-infra image directly. For the monorepo, we use a different approach to build the kubekins-with-kind image. It would be beneficial if we can consolidate the OSS one with the monorepo one.

nan-yu: I see where the confusion is coming from now.
Previously, the `kubekins-with-kind` image has `kind` embeded. It makes sense to host the customized image in a registry we control. Now, https://github.com/GoogleContainerTools/kpt-config-sync/pull/781 changes it to install kind at runtime, the kubukins image is no longer customized. I guess at this point, we can use the k8s image directly.

If we use it directly, we only need to update the image version here once. Otherwise, we need the following steps.
1. update the k8s image in https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/build/prow/kubekins-e2e/Dockerfile.
2. run `build-kubekins-e2e`: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/Makefile.oss.prow#L29
3. run `push-kubekins-e2e`: https://github.com/GoogleContainerTools/kpt-config-sync/blob/main/Makefile.oss.prow#L37
4. update the test image used in the presubmit prow job here

sdowell: I think the reasons I mentioned above are still motivations for maintaining our own Dockerfile/image, even if it is currently just a retag. I considered adding a postsubmit job to automate publishing the images, but it didn't seem like too much extra effort to manually publish them once in a while - given we don't make changes very often

PR #2005, opened 2023-07-17
Use the absolute path to shell

koln67:
The path to shell requires the / to be accessed

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
@@ -9,7 +9,7 @@ presubmits:
     decorate: true
     spec:
       containers:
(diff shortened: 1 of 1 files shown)


dorileo:
as we talked offline this PR (and the previous one adding this) is wrong. adding a hold to make sure it doesn't get merged.

prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
```diff
@@ -12,7 +12,7 @@ presubmits:
       - image: gcr.io/compute-image-tools/daisy:latest
```
dorileo, line 12: I guess we also want to change the image, right?

koln67: done

PR #2011, opened 2023-07-18
Run bash and daisy in one command

koln67:
Set the secret variable before running daisy, all in one bash command

### DIFF
--- a/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
+++ b/prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
@@ -15,12 +15,7 @@ presubmits:
         - "bash"
         args:
         - "-c"
-        - "sbom_root=$(gcloud secrets versions access latest --secret=sbom-util-secret --project=gcp-guest); "
-        - "/daisy"
-        - "project=compute-image-test-pool-001" 
-        - "var:sbom_util_gcs_root=$sbom_root"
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleCloudPlatform/gcp-guest/compute-image-tools.yaml
```diff
@@ -15,11 +15,7 @@ presubmits:
         - "bash"
         args:
         - "-c"
-        - "sbom_root=$(gcloud secrets versions access latest --secret=sbom-util-secret --project=gcp-guest); "
-        - "/daisy"
-        - "project=compute-image-test-pool-001" 
-        - "var:sbom_util_gcs_root=$sbom_root"
-        - "daisy_workflows/sbom_validation/enterprise_sbom_test.wf.json"
+        - "sbom_root=$(gcloud secrets versions access latest --secret=sbom-util-secret --project=gcp-guest); /daisy -project=studious-sign-357617 -var:sbom_util_gcs_root=$sbom_root daisy_workflows/sbom_validation/enterprise_sbom_test.wf.json"
```
Quintonamore, line 18: I've never heard of project `studious-sign-357617` why are we using that?

koln67: fixed, that was what I was using when testing this manually but it should be compute-image-test-pool-001

PR #2019, opened 2023-07-27
increase cluster size for stress tests

sdowell:
The stress tests manage and then delete a large number of resources. The reset code currently times out with the given cluster shape. Increasing the number of nodes enables the stress test to complete within the default timeout (experimentally).


nan-yu (approved):
/lgtm

PR #2035, opened 2023-08-17
remove release branch periodics

sdowell:
We are about to move the release branch infra forward for v1.16, with which we will be using the new job shape that leverages cluster pooling. The old infra needs to be torn down so that the new jobs can be set up.


nan-yu (approved):
/lgtm

PR #2062, opened 2023-09-12
Add new test suites to 12 hour periodic

koln67:
Add hotattach and shape validation tests to the test grid.

Currently the test grid does not run any windows images: we can consider adding that in the future as well.


a-crate:
Most test projects still don't have capacity to run shapevaldation, there is a request in to get enough capacity to run it in compute-image-test-pool-005 in us-central1-a.  We would need to put another capacity increase request in to accommodate the concourse jobs and addition to the testgrid jobs, or just request the same capacity in a different test pool and segregate them.

PR #2063, opened 2023-09-13
add gke-gcloud-auth-plugin kubeconfig file for the oss-prow

listx:
The oss-prow cluster had its CA root certificate (public key) rotated earlier this week because it was going to expire next month. However this means that our Prow components need to update their kubeconfigs to schedule trusted jobs (e.g., Prow Controller Manager needs to talk to the K8s API server on the oss-prow cluster), because they still use the old kubeconfig for the oss-prow cluster with the old CA root certificate.

Check in the minimal kubeconfig for access into the oss-prow cluster. This minimal configuration was created as follows:

First we ran

    gcloud --project=oss-prow container clusters get-credentials prow --region us-west1-a

which fetched the correct cluster IP (which was rotated as part of the CA rotation) and CA certificate (public key), but which uses the new gke-gcloud-auth-plugin to obtain client credentials. The kubeconfigs.yaml content was the result of running

    kubectl config view --context=oss -oyaml >kubeconfigs.yaml

followed by removing all other non-oss-prow cluster entries. The "certificate-authority-data" field was not exported with the above command (it just had "DATA+OMITTED"), so we had to look it up directly in the raw "~/.kube/config" file manually. This file was tested with

    kubectl --kubeconfig ./kubeconfigs.yaml get po

to check that it worked with the oss-prow cluster.

As part of this change, also create a Makefile.gcloud.mk rule to update this kubeconfigs.yaml file going forward (we might want to add in kubeconfigs for other clusters here in the future). The rule here is essentially copied from the update-config rule in the main Makefile. Note that we must first run

    kubectl create configmap kubeconfigs --from-file=kubeconfigs.yaml=kubeconfigs.yaml

in order to first create the configmap in the cluster, after which the

    ... | kubectl replace configmap kubeconfigs -f -

will work (otherwise this part will fail because the "kubeconfigs" configmap doesn't exist yet).

A subsequent change will mount the configmap into our components (such as Prow Controller Manager) that need it.

/cc @michelle192837 @cjwagner @airbornepony @timwangmusic


listx:
/hold

@cjwagner How come we don't run the `update-config` or `update-plugins` make targets in https://github.com/GoogleCloudPlatform/oss-test-infra/blob/4c2687926851a5ce12e6a53703b0ab4720015758/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml#L3-L25 ? (Did we purposely leave these targets out for some reason, so that we have to run them manually any time there's a change to config.yaml or plugins.yaml?)

michelle192837 (approved):

prow/oss/kubeconfigs.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: gke_oss-prow_us-west1-a_prow
+current-context: gke_oss-prow_us-west1-a_prow
+kind: Config
+preferences: {}
+users:
+- name: gke_oss-prow_us-west1-a_prow
+  user:
+    exec:
+      apiVersion: client.authentication.k8s.io/v1beta1
+      args: null
```
listx, line 20: Need the application default credentials flag here `--use_application_default_credentials`

listx: Done in #2065

prow/Makefile.gcloud.mk
```diff
@@ -39,3 +39,7 @@ get-cluster-credentials: save-kubeconfig activate-serviceaccount
 .PHONY: get-build-cluster-credentials
 get-build-cluster-credentials: save-kubeconfig activate-serviceaccount
 	gcloud container clusters get-credentials "$(CLUSTER_BUILD)" --project="$(PROJECT_BUILD)" --zone="$(ZONE)"
+
+.PHONY: update-kubeconfigs
+update-kubeconfigs:
+	kubectl create configmap kubeconfigs --from-file=kubeconfigs.yaml=kubeconfigs.yaml --dry-run=client -o yaml | kubectl replace configmap kubeconfigs -f -
```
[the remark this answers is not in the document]
listx: Done (we use `kubectl apply` now) in #2065

PR #2065, opened 2023-09-13
fix kubeconfig for test-infra-trusted cluster

listx:
The update to the kubeconfigs.yaml file has already been applied to the oss-prow cluster, and is a NOP.

/cc @michelle192837 @cjwagner @airbornepony @timwangmusic

PR #2066, opened 2023-09-13
mount kubeconfigs configmap into other prow components

listx:
We've already tested that the settings in
https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2065 work for prow-controller-manager.

Mount the kubeconfig into crier, deck, deck-blueprints (for consistency with deck), hook, and sinker.

/cc @michelle192837 @cjwagner @airbornepony @timwangmusic


timwangmusic (approved):
LGTM

PR #2072, opened 2023-09-19
Add windows and partner images to CIT performance periodics

a-crate:


prow/prowjobs/GoogleCloudPlatform/gcp-guest/gcp-guest-config.yaml
```diff
@@ -426,6 +426,6 @@ periodics:
...
...jects/ubuntu-os-cloud/global/images/family/ubuntu-minimal-2204-lts,projects/ubuntu-os-cloud/global/images/family/ubuntu-minimal-2304-amd64,projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-1604-lts,projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-1804-lts,projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-2004-lts,projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-2204-lts,projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-fips-1804-lts,projects/ubuntu-os-pro-cloud/global/images/family/ubuntu-pro-fips-2004-lts,projects/windows-cloud/global/images/family/windows-2012-r2,projects/windows-cloud/global/images/family/windows-2012-r2-core,projects/windows-cloud/global/images/family/windows-2016,projects/windows-cloud/global/images/family/windows-2016-core,projects/windows-cloud/global/images/family/windows-2019,projects/windows-cloud/global/images/family/windows-2019-core,projects/windows-cloud/global/images/family/windows-2019-core-for-containers,projects/windows-cloud/global/images/family/windows-2019-for-containers,projects/windows-cloud/global/images/family/windows-2022,projects/windows-cloud/global/images/family/windows-2022-core"
```
zmarano, line 429: There are two possibilities for sourcing some of the images. `bct-prod-images` is our to be released project, vs the public projects which are released images. I think for the sake of sanity right now, we should use the public projects in this testgrid (so `windows-cloud` is correct. But then we should also change the projects for Debian and the EL distros to their respective public projects.

a-crate: Should we switch these for the regular periodics as well? Or leave that as bct-prod-images?

jjerger: Images staged in bct-prod-images will already have needed to pass CIT tests on build, right? I think if that already covers the use cases we want we could swap to public images.

I think it comes down to if we want the periodics to signal potential issues already out in the wild, or catch them in a pre-release phase

a-crate: IIUC the test gets run after the publish step, so we can end up with failing images in bct-prod-images but we will still have the failure signal in concourse.

zmarano: At the moment- the performance periodics are useful to signal against changes in infrastructure against the OS. Eventually, when that is stable enough, we would also want to look for changes in the OS against the infrastructure. We could make the same case for the regular periodic tests- because we are also validating them prior to release.

a-crate: As long as we're updating the images here I'm going to add debian-12 now that that's released.

PR #2088, opened 2023-10-02
Add more relevant OSes to the periodics suite

zmarano:
* Change to 2 hour intervals
* Use us-central1-c as the default zone
* Add hot attach test


a-crate:
If hotattach is part of regular periodics, do we need it in performance periodics too?
/lgtm

zmarano:
> If hotattach is part of regular periodics, do we need it in performance periodics too? /lgtm

Oh definitely not (it's not a performance test). PR incoming.

PR #2116, opened 2023-11-10
Revamp guest-test-infra presubmit checks

a-crate:
We are missing a lot of gobuild/gocheck/gotest coverage because these containers are scoped to only subdirectories which belong to the same module as the main directory. This change will add each module as its own presubmit directory to ensure coverage.

citpresubmit still handles building and testing because of the way we handle test code in those directories.


a-crate:
/hold
Should be merged at the same time as https://github.com/GoogleCloudPlatform/guest-test-infra/pull/916

PR #2128, opened 2023-11-29
kes-kicker: target 'default' namespace

listx:
In #2125, we created various resources (ServiceAccount, Role,
RoleBinding, CronJob) all with the name "kes-kicker". Looking at the
cluster after these configurations have been applied, they all appear to
be in the "test-pods" namespace.

Consequently, we get an error like this in the logs from the kes-kicker
pod when it gets scheduled to satisfy the CronJob definition:

    Error from server (NotFound): deployments.apps "kubernetes-external-secrets" not found

This appears to be because the kubectl command inside kes-kicker is
operating inside the test-pods namespace.

Recreate the same resources in the default namespace. But also, give the
kubectl invocation an namespace because it's more strict that way and it
can't hurt.

/cc @cjwagner @airbornepony @timwangmusic

### DIFF
--- a/prow/oss/cluster/kes-kicker.yaml
+++ b/prow/oss/cluster/kes-kicker.yaml
@@ -2,11 +2,13 @@ kind: ServiceAccount
 apiVersion: v1
 metadata:
   name: kes-kicker
+  namespace: default
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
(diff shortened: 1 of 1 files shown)


cjwagner (approved):
LGTM

listx:
I've cleaned up (deleted) the resources created in #2125 in the test-pods namespace.

PR #2131, opened 2023-12-05
Offboard googleforgames{,/agones}

zmerlynn:
We are not using Prow anymore and are working on disabling the webhooks.


roberthbailey:
LGTM

chases2:
The webhooks have been removed, so this should be good to go

/approve
/lgtm

PR #2142, opened 2023-12-16
Revert "make `kfp-kubernetes` execution tests required"

connor-mccarthy:
Tests were passing, but are now failing. Pipelines are stuck in a "Running" state, despite inner tasks all succeeding.

image

/assign @chensun


chensun (approved):
/lgtm

PR #2144, opened 2023-12-18
build & test root, registry image, cleanerupper, and cit whenever go.mod or go.sum is changed

a-crate:
/cc @elicriffield @zmarano 

the cit gobuild config already does this


a-crate:
Ping

PR #2174, opened 2024-01-25
Exclude all 4th gen machine types from perf dashboard

a-crate:
The last two runs are timing out because the individual test cases are never finishing. Daisy isn't logging anything in the storage buckets, the only thing uploaded is the test binaries and wrapper. I have no idea what's causing this but it's definitely related to the 4th gen machine types, I want to exclude them for now until figuring out a root cause.

/cc @elicriffield @drewhli


a-crate:
Going to reopen this actually, there's some more issues with 4th gen machine type perf testing that I'm still debugging
/cc @koln67

PR #2176, opened 2024-01-26
Bump timeout and parallel count for cit-periodics

a-crate:
Latest runs are still timing out
/cc @zmarano @koln67


a-crate:
Merged but didn't actually merge? Weird. I'll close it since the commits are in the repo.

PR #2182, opened 2024-02-02
k8s-config-connector: override plugin

acpana:
add override plugin to k8s-config-connector


cjwagner (approved):
LGTM, unhold when ready.

acpana:
cc @yuwenma we can re-open this if we ever need it but we haven't done an override with a slash command in the last yr. Would be nice to have it but closing for now.

PR #2186, opened 2024-02-07
ci: add helm/oci provider options for CS e2e

sdowell:
This adds the new helm/oci provider flags for the periodic job configs. This sets the GKE tests to run with the artifact registry provider and the kind tests to run with the local provider.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
@@ -45,6 +45,10 @@ prow_ignored:
         value: "pd-ssd"
       - name: GKE_DISK_SIZE
         value: "50Gb"
+      - name: E2E_OCI_PROVIDER
+        value: "gar"
+      - name: E2E_HELM_PROVIDER
+        value: "gar"
       resources:
         requests:
           memory: "8Gi"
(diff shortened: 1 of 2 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
```diff
@@ -75,6 +79,8 @@ periodics:
       args:
       - make
       - test-e2e-kind-multi-repo
+      - 'E2E_OCI_PROVIDER=local'
```
nan-yu, line 82: Do we need to set it explicitly? `local` is the default value.

sdowell: We need to override the value set in the parent template, which is set to `gar` so that it applies to all of the other periodic jobs (which run on GKE)

nan-yu: ACK

nan-yu (approved):
/lgtm

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
sdowell: Alternatively we could update the `test-e2e-kind-multi-repo` target in the repo to always set the providers to local

PR #2194, opened 2024-02-14
ci: stagger execution of CS periodics

sdowell:
We are running into some quota issues with the number of concurrent tests running at once. This change reduces the test frequency to every two hours, and staggers the execution such that the standard jobs run on even hours and the autopilot jobs run on odd hours.


karlkfi (approved):
/lgtm

PR #2202, opened 2024-02-21
ci: further stagger each large GKE job

sdowell:
This change gives a minute offset to each of the big jobs to further stagger them. This is primarily motivated by the autopilot jobs so that not all clusters are scaling from zero at the same time.


karlkfi (approved):
/lgtm

PR #2205, opened 2024-02-22
Update kubeflow-pipelines-presubmits.yaml to use go 1.21 for backend presubmit

chensun:


Tomcli (commented):
thanks @chensun 
/lgtm

PR #2209, opened 2024-02-26
configsync: increase stress test node size

karlkfi:
Use fewer larger nodes, to give the reconcilers more headroom to run faster, even with low resource requests.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -344,7 +344,7 @@ periodics:
       - 'GKE_CLUSTER_VERSION=latest'
       - 'E2E_CLUSTER_PREFIX=standard-rapid-latest-stress'
       - 'E2E_NUM_CLUSTERS=1'
-      - 'GKE_NUM_NODES=3' # stress test needs a bigger cluster to handle finalizing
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -344,7 +344,7 @@ periodics:
       - 'GKE_CLUSTER_VERSION=latest'
       - 'E2E_CLUSTER_PREFIX=standard-rapid-latest-stress'
       - 'E2E_NUM_CLUSTERS=1'
-      - 'GKE_NUM_NODES=3' # stress test needs a bigger cluster to handle finalizing
+      - 'GKE_MACHINE_TYPE=n2-standard-8' # stress test uses bigger nodes to run more quickly
```
sdowell, line 347: To be equivalent to the previous config, this should specify both `GKE_NUM_NODES=3` and `GKE_MACHINE_TYPE=n2-standard-8`

karlkfi: GKE_NUM_NODES=1 with regional means 1 per zone, so 3 total.

sdowell: Although maybe that won't be necessary since we switched from zonal to regional

sdowell (approved):

PR #2215, opened 2024-03-07
Set up CIT alerting

a-crate:
Bug guest os images on alert
For cit-periodics: alert when 1.2% of cit tests are failing 3 times in a row. We run 4x a day so this should be enough to alert if CIT is consistently failing many tests for a whole day. 1.2% is enough to pass if all el7 networking tests fail, plus a small margin for a few flaky individual tests.

For oslogin-periodics: alert when 1% of tests are failing. Less than 4 failing tests basically, which is a number that we've been hitting extremely consistently since adding more test users. Alert on 2 failures, and drop the interval to 12 hours so we're running this twice a day.


elicriffield (approved):
LGTM

PR #2220, opened 2024-03-11
Only configure alerting thresholds on prow annotations

a-crate:
Also disable setting exit status to avoid always marking the job failed.


a-crate:
moved email and debug url to testgrid itself: kubernetes/test-infra#32216
/cc @elicriffield

elicriffield (approved):
LGTM

PR #2228, opened 2024-03-20
Change the default Git provider to CSR

nan-yu:
The kind test covers testing with the local test-git-server. This commit
changes other CI jobs to test with the Cloud Source Repository so that
we don't need a separate job for Git + WI tests.

It also removes the test regex for those one-off jobs because
https://github.com/GoogleContainerTools/kpt-config-sync/pull/1160
changes the e2e test to run these tests if and only if the flag is
specified.
    
b/325119647

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -60,6 +60,8 @@ prow_ignored:
         value: "pd-ssd"
       - name: GKE_DISK_SIZE
         value: "25Gb"
+      - name: E2E_GIT_PROVIDER
+        value: "csr"
       - name: E2E_OCI_PROVIDER
         value: "gar"
       - name: E2E_HELM_PROVIDER
@@ -93,6 +95,8 @@ prow_ignored:
       # This can be set to true to destroy clusters after test execution
       - name: E2E_DESTROY_CLUSTERS
         value: "false"
+      - name: E2E_GIT_PROVIDER
+        value: "csr"
       - name: E2E_OCI_PROVIDER
         value: "gar"
       - name: E2E_HELM_PROVIDER
@@ -125,6 +129,7 @@ periodics:
       args:
       - make
       - test-e2e-kind-multi-repo
+      - 'E2E_GIT_PROVIDER=local'
       - 'E2E_OCI_PROVIDER=local'
       - 'E2E_HELM_PROVIDER=local'
       securityContext:
@@ -212,7 +217,7 @@ periodics:
       - 'GCP_SUBNETWORK=prow-e2e-subnetwork-usw1-6'
       - 'GKE_RELEASE_CHANNEL=regular'
       - 'E2E_CLUSTER_PREFIX=standard-regular-bitbucket'
-      - 'E2E_ARGS=--git-provider=bitbucket'
+      - 'E2E_GIT_PROVIDER=bitbucket'
 
 - <<: *config-sync-standard-job
   name: kpt-config-sync-standard-regular-gitlab
@@ -227,7 +232,7 @@ periodics:
       - 'GCP_SUBNETWORK=prow-e2e-subnetwork-usw1-7'
       - 'GKE_RELEASE_CHANNEL=regular'
       - 'E2E_CLUSTER_PREFIX=standard-regular-gitlab'
-      - 'E2E_ARGS=--git-provider=gitlab'
+      - 'E2E_GIT_PROVIDER=gitlab'
 
 #### End git provider specific jobs
 #### Begin GKE autopilot jobs
@@ -311,7 +316,7 @@ periodics:
       - 'GKE_RELEASE_CHANNEL=regular'
       - 'E2E_CLUSTER_PREFIX=standard-regular-kcc'
       - 'E2E_NUM_CLUSTERS=1'
-      - 'E2E_ARGS=--kcc -run=TestKCC*'
+      - 'E2E_ARGS=--kcc'
 
 - <<: *config-sync-standard-job
   name: kpt-config-sync-standard-regular-gcenode
@@ -327,7 +332,7 @@ periodics:
       - 'GKE_RELEASE_CHANNEL=regular'
       - 'E2E_CLUSTER_PREFIX=standard-regular-gcenode'
       - 'E2E_NUM_CLUSTERS=1'
-      - 'E2E_ARGS=--gcenode -run=TestGCENode'
+      - 'E2E_ARGS=--gcenode'
 
 - <<: *config-sync-standard-job
   name: kpt-config-sync-standard-rapid-latest-stress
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
```diff
@@ -313,7 +313,29 @@ periodics:
       - 'GKE_RELEASE_CHANNEL=regular'
       - 'E2E_CLUSTER_PREFIX=standard-regular-gcenode'
       - 'E2E_NUM_CLUSTERS=1'
-      - 'E2E_ARGS=--gcenode -run=TestGCENode'
+      - 'E2E_GIT_PROVIDER=csr'
```
sdowell, line 316: AFAIK we're not ready to port these new configs to the release CI, I think this should be done after 1.18 is cut

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -327,7 +327,27 @@ periodics:
       - 'GKE_RELEASE_CHANNEL=regular'
       - 'E2E_CLUSTER_PREFIX=standard-regular-gcenode'
       - 'E2E_NUM_CLUSTERS=1'
-      - 'E2E_ARGS=--gcenode -run=TestGCENode'
+      - 'E2E_GIT_PROVIDER=csr'
+      - 'E2E_ARGS=--gcenode -run=TestGCENode*'
```
sdowell, line 331: The `-run` regex is no longer needed, right?

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -327,7 +327,27 @@ periodics:
...
+  spec:
+    <<: *config-sync-standard-job-spec
+    containers:
+    - <<: *config-sync-standard-job-container
+      args:
+      - 'GCP_SUBNETWORK=prow-e2e-subnetwork-usw1-11'
+      - 'GKE_RELEASE_CHANNEL=regular'
+      - 'E2E_CLUSTER_PREFIX=standard-regular-workloadidentity'
+      - 'E2E_NUM_CLUSTERS=1'
+      - 'E2E_GIT_PROVIDER=csr'  # run CSR+WI tests on a separate job
```
sdowell, line 347: Should we consider defaulting to `csr` instead of `local` for the other jobs?

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
nan-yu: It is not required, but having the regex here makes the testgrid cleaner. 
With the regex, the testgrid only shows the selected tests. Otherwise, it shows other tests as skipped.

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
nan-yu: I don't think all tests in other jobs can run with CSR, for example, proxy test, ca_cert test, and etc.
If they have to run with the local test-git-server, we should default to `local`.

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
nan-yu: Ah, right. Thanks for catching this. Reverted.

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
sdowell: It seems like it could be a gotcha, given that with the "if and only if" logic a developer might assume they just need to set the `GCENodeTest` option and the test will be run. But it reality, they need to both set the option and have a specific name format for the test itself.

It also seems okay to display the skipped tests in the testgrid, but I don't feel too strongly

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
sdowell: It looks like it's only the `caCert` and `knownHosts` tests which require local git provider, which adds up to 5 tests. Either way we are skipping some tests, and note that the kind cluster tests (both presubmit and periodic) will run with `local` git provider to exercise `caCert`/`knownHosts`. I also feel that test coverage for CSR may be more meaningful than our bespoke local git provider.

We discussed this offline, but it still feels like a bit of an inconsistency that we run the full test suite for gitlab/bitbucket but only this small subset of coverage for CSR.

Let me know what you think, I'm fine with this if you disagree

nan-yu: I like the idea of having more test coverage for CSR. If we make CSR the default git-server, then we don't need a separate job to run the GIT+WI tests.
My two concerns are:
1. It may take longer for each test to provision a  CSR repo and push commits than using the local test-git-server, but I'm not sure how much longer it is. 
2. We need to add a `ntopts.RequireLocalGitServer` to filter out those tests that require `local` git provider,  including `no_ssl_verify_test.go`, `private_cert_secret_test.go`, `proxy_test.go`, and `ssh_known_hosts_test.go`. I'm not too concerned since they're covered by the kind cluster tests.

That being said, it may be worth to exercise the change and check the overall test duration.

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
nan-yu: SG. I agree with additional regex adds an assumption on the test names.

The testgrid also providers an option to exclude filter by regex. We can use that to filter out the skipped test.

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
nan-yu: Nevermind, `private_cert_secret_test.go` and `ssh_known_hosts_test.go` already have `ntopts.RequireLocalGitProvider`.
`proxy_test.go` sets up the http proxy with a GitHub repo, and `no_ssl_verify_test.go` verifies if the env var is configured.

This PR is updated to change the default git provider to CSR.

prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
nan-yu: done

sdowell (approved):
/lgtm

PR #2230, opened 2024-03-25
Run performance periodics on a longer, lower interval

a-crate:
This should help with quota issues and be a more efficient use of resources. The images in cit-performance-periodics change very infrequently so the weekly interval should be fine for testing. The images in ubuntu-daily-cit-performance change more frequently so I lowered to half a week. I have more detailed math if you want to check yourself but based on historical data I'm expecting runtimes of 8.4 - 19.7 hours (with an average of 11.4 hours), which gives a four hour buffer before timeouts. parallel_count=1 is more efficient but doubles these runtimes and running a prow job for longer than 24 hours requires some extra configuration beyond just time IIUC so I've stuck with this for now.

/cc @bkatyl @elicriffield


elicriffield (approved):
LGTM

PR #2239, opened 2024-04-03
Fix gcp-guest CA cert and IP address.

cjwagner:
This has already been deployed.
/assign @airbornepony


airbornepony (approved):
/hold for author

PR #2267, opened 2024-05-03
Change the default Git provider to CSR

nan-yu:
This commit brings changes from #2228 and #2233 to the release branch.


sdowell (approved):
/lgtm

PR #2271, opened 2024-05-06
Require self-approval for Kubeflow repos

thesuperzapper:
As discussed in https://github.com/kubeflow/kubeflow/issues/7549#issuecomment-2091871635, and other places.

We need to ensure that approvers are explicitly approving their PRs in the Kubeflow org, as this prevents people from accidentally merging PRs with a single `/lgtm`.

This PR updates the prow configs so that all currently active Kubeflow repos require people who are approvers to explicitly `/approve` their own PRs, if that is their intention.

__NOTE:__ it might be possible to just enable this at the org level by including `kubeflow` as a list element, but I am not super familiar with how the config works.

### DIFF
--- a/prow/oss/plugins.yaml
+++ b/prow/oss/plugins.yaml
@@ -39,19 +39,25 @@ approve:
   require_self_approval: false
   commandHelpLink: https://oss.gprow.dev/command-help
 - repos:
-  - GoogleContainerTools
-  - kubeflow/caffe2-operator
-  - kubeflow/chainer-operator
-  - kubeflow/common
-  - kubeflow/fate-operator
-  - kubeflow/kubeflow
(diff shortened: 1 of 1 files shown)


michelle192837 (approved):
/hold

Unhold at will, holding in case you want others to also approve this first.

andreyvelich (commented):
Thanks for the update @thesuperzapper 
/lgtm
Please take a look.
/assign @zijianjoy @juliusvonkohout @jbottum @terrytangyuan @kimwnasptd

terrytangyuan (commented):
/lgtm

thesuperzapper:
Pretty sure I cant unhold, but I will try.

/unhold

PR #2279, opened 2024-05-10
Enable trigger for cloud-image-tests

a-crate:


a-crate:
Looks like it's enabled by default. Sorry for noise

PR #2281, opened 2024-05-15
Pin to a specific cluster version for Autopilot rapid and rapid-latest

nan-yu:
The current Autopilot jobs failed due to Cilium API client timeout error.
The GKE team has marked a few versions as bad. This commit pins to a specific working version for the rapid and rapid-latest CIs. The regular channel doesn't have the fix available yet. It may take another 1-2 weeks until the fix is available in the regular channel. The stable channel seems to be okay with an older version.

b/339056595


sdowell (approved):
/lgtm

feel free to unhold when ready

nan-yu:
`1.29.4-gke.1670000` have been fully rolled out. 

/unhold

PR #2282, opened 2024-05-15
ci: use local git provider as default for CS jobs

sdowell:
With the number of concurrent jobs we have running against a single project, we were frequently hitting the CSR api quota. This switches back to the local git provider as the default and creates a separate job for testing CSR.


nan-yu (approved):
/lgtm

PR #2284, opened 2024-05-16
chore: Bump python version for Kubeflow Pipelines CI

gmfrasca:
- Python 3.7 is in EOL, so bump containers used in KFP CI to python 3.8 for now
- Additionally, the KFP SDK no longer supports python3.7 so tests using these images are broken


gmfrasca:
Just for traceability, several tests started breaking due to dropping python3.7 from the KFP SDK, which was done here: https://github.com/kubeflow/pipelines/pull/10750, 

prow logs showing this error on `kubeflow-pipelines-samples-v2` check, for example:
```
ERROR: Package 'kfp' requires a different Python: 3.7.17 not in '<3.13.0,>=3.8.0'
```

chensun (approved):
/lgtm
/approve

PR #2296, opened 2024-06-05
chore: Remove KFP presubmit backend test prow config

DharmitD:
As part of migration from Prow to GH Action Workflows for KFP, we have a PR proposed to migrate presubmit backend tests to a GHA: https://github.com/kubeflow/pipelines/pull/10871

This PR removes presubmit backend tests from the prow config in parallel.


chensun (approved):
/lgtm
/approve

Thanks!

PR #2297, opened 2024-06-10
Add oldest images from each server family to package test

jjerger:
Tests for regressions between current image and first images


jjerger:
Brian - with the deprecation of the windows-2012-r2 image family all the images were marked obsolete so those images are not usable, however the windows-2012-r2-internal images are still accessible so I added those in. I'm not sure how different those would be from the obsoleted family so I wanted to bring this to your attention before merging.

bkatyl (approved):

PR #2319, opened 2024-07-09
chore: use renamed config-sync test targets

sdowell:
The multi-repo aliases reflect a legacy naming scheme, and all current branches support the new name.


nan-yu (approved):
/lgtm

PR #2323, opened 2024-07-17
enable MTU workaround for dind presets

sdowell:
This fixes an issue where the dind container is on a host with a MTU lower than the typical default of 1500, resulting in a mismatch.

Upstream maintainers are considering making this the default behavior of the script, but for now this is a quick workaround.

See: https://github.com/kubernetes/test-infra/blob/94f76aedb29b50131fe5dc5e131adcd23f6d4f22/images/bootstrap/runner.sh#L87

### DIFF
--- a/prow/oss/config.yaml
+++ b/prow/oss/config.yaml
@@ -298,6 +298,8 @@ presets:
   env:
   - name: DOCKER_IN_DOCKER_ENABLED
     value: "true"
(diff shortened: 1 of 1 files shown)


BenTheElder (commented):
/lgtm

TODO: consider making this the default in the script (or something more elaborate that detects MTU and reconfigures docker instead of the iptables hack)

michelle192837 (approved):

PR #2324, opened 2024-07-17
add node label selectors to all CS jobs

sdowell:
This is more flexible than the strict gke node pool selector, allowing the underlying node pools to be changed without downtime and/or coordinating changes to the prow config. Also separates presubmits from periodic nodes for better isolation/security.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
@@ -25,7 +25,8 @@ prow_ignored:
           cpu: "4000m"
     nodeSelector:
       # This job requires 8vCPUs or less, so it is "small".
-      cloud.google.com/gke-nodepool: small-job-pool
+      kpt-config-sync/type: periodic
+      kpt-config-sync/size: small
 # Base job definition for standard jobs
 - &config-sync-standard-job
   <<: *config-sync-base-job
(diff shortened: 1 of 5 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
```diff
@@ -25,7 +25,8 @@ prow_ignored:
           cpu: "4000m"
     nodeSelector:
       # This job requires 8vCPUs or less, so it is "small".
-      cloud.google.com/gke-nodepool: small-job-pool
+      kpt-config-sync/type: periodic
```
nan-yu, line 28: Are we labeling the CI nodes right now?

sdowell: We're using these labels for the nodes that the kind presubmits run on but not the other nodes yet. I'm adding labels to the other node pools

nan-yu: ACK. We would need this to be documented in the playbook in case of recreating those nodepools.

sdowell: Sure, I can add a playbook

nan-yu (approved):
/lgtm

PR #2325, opened 2024-07-18
chore: sync kind periodics with presubmits

sdowell:
The kind periodic config has drifted a bit from the presubmit config. Updating them to be in sync again.


nan-yu (approved):
/lgtm

PR #2337, opened 2024-08-02
ci: add prowjob for Config Sync githubapp

sdowell:
The githubapp tests are currently implemented as a one off test suite similar to kcc and gcenode. This creates a separate prowjob which runs only the new githubapp tests.

Depends on https://github.com/GoogleContainerTools/kpt-config-sync/pull/1372


sdowell:
Hold until Config Sync PR is submitted
/hold

nan-yu (approved):
/lgtm

PR #2341, opened 2024-08-06
Update image locations to use new locations.

michelle192837:
This includes three changes:
- Upstream misc test-infra images are now published to gcr.io/k8s-staging-test-infra, so update image references to use that. gcr.io/k8s-prow > gcr.io/k8s-staging-test-infra
- Upstream Prow images use the gob-prow AR images that are maintained by the Prow dev team (and are separate from the ones published in k8s-sigs/prow). gcr.io/k8s-prow > us-central1-docker.pkg.dev/gob-prow/prow-images
- Autobump config needs to use the new locations, and has a split now to update the two different locations (for Prow and misc images). Prow images remain consistent, misc images don't have to.

Images in gcr.io/k8s-staging-test-infra are public, but this requires that oss-prow has permissions to pull images from us-central1-docker.pkg.dev/gob-prow/prow-images. (I've granted permissions for oss-prow's default SA to view Artifact Registry in gob-prow, which I _think_ should take care of it.)

### DIFF
--- a/prow/bump.sh
+++ b/prow/bump.sh
@@ -30,17 +30,6 @@ if ! (${SED} --version 2>&1 | grep -q GNU); then
   exit 1
 fi
 
-TAC=tac
-
-if command -v gtac &>/dev/null; then
-  TAC=gtac
-fi
-
-if ! command -v "${TAC}" &>/dev/null; then
-  echo "tac (reverse cat) required. If on OS X then 'brew install coreutils'." >&2
-  exit 1
-fi
-
 cd "$(git rev-parse --show-toplevel)"
 
 usage() {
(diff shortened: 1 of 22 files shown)


michelle192837:
Updated reviewers. (Cole has context on the locations we're changing this to, and Sean your review is optional, but I know you saw some of the image location changes upstream).

cjwagner (approved):
LGTM. Lets have Simon take a quick look too though.
/hold

simony-gke:
/lgtm

Please make sure bump.sh can access us-central1-docker.pkg.dev/gob-prow/prow-images.

michelle192837:
Update bump.sh as well as the autobump config (I had local changes that didn't get saved).

prow/bump.sh
```diff
@@ -30,17 +30,6 @@ if ! (${SED} --version 2>&1 | grep -q GNU); then
   exit 1
```
cjwagner, line 30: Is bump.sh still being used somewhere? It looks like this prow instance is bumped using the generic-autobumper instead: https://github.com/GoogleCloudPlatform/oss-test-infra/blob/0412bc0d9d15f2e1d1f27bbbb256950d4f0a0d07/prow/prowjobs/GoogleCloudPlatform/oss-test-infra/gcp-oss-test-infra-config.yaml#L81-L153

I suspect this script and the reference to it in the README.md are outdated.

michelle192837: I think it's only for manual updates. It was working before I made the change, so I think fixing it is legit enough, but I'm also okay to just drop it from the repo. (It's referencing a few very specific files anyways and doesn't cover all of what the autobump does).

cjwagner: /shrug
SGTM
/meowvie tools

cjwagner (approved):

PR #2355, opened 2024-08-29
minimize concurrent AR repo creates/deletes

sdowell:
AR has a fixed quota of 30 repo create/delete requests per minute, which is occasionally being exhausted in the current configuration. This switches the gitprovider job types to use local helm/oci provider so that they do not create AR repos, given that standard-regular already has test coverage for AR.

Also sets the cron schedule for standard/autopilot regular to be on the same even/odd hour schedule as the other standard/autopilot jobs.


nan-yu:
The kcc/gcenode/githubapp  jobs can also use local helm/oci provider.

nan-yu (approved):
/lgtm

PR #2356, opened 2024-08-29
Bump the Config Sync release branch to v1.19

nan-yu:
b/354790801


sdowell:
> /hold It needs to wait until [feat: add github app auth to git-sync api](https://github.com/GoogleContainerTools/kpt-config-sync/pull/1372) is cherrypicked to the release branch.

I think we could just merge this now and let the job fail until we cherry pick githubapp. Alternatively we could comment out the job and add it in after the cherry pick. Either way I think it would be preferable to get this change in asap

sdowell (approved):
/lgtm

nan-yu:
> I think we could just merge this now and let the job fail until we cherry pick githubapp. Alternatively we could comment out the job and add it in after the cherry pick. Either way I think it would be preferable to get this change in asap

SGTM. The Config Sync 1.19 release branch is being created. Once it is done, we can do the cherrypick.

/unhold

PR #2364, opened 2024-09-03
Add cron jobs to restart pcm/crier/sinker/deck every day

nathanperkins:
This resolves an issue where components cannot perform their tasks on cluster with legacy auth due to expired credentials, resulting in authentication / authorization issues. For example, we were seeing pod creation issues from the prow-controller-manager.


cjwagner (approved):
/retest

PR #2367, opened 2024-09-04
chore: remove unused envs from cs kind presubmit

sdowell:
The kind presubmits do not require a GCP project, and these arguments are unused.


nan-yu (approved):
/lgtm

PR #2370, opened 2024-09-06
create new clusters for autopilot-rapid tests

sdowell:
There is an issue with the cluster autoscaler which is currently undergoing a RCA. Changing the cluster names will allow new clusters to be created and for the tests to pass in the meantime.


nan-yu (approved):
/lgtm

PR #2377, opened 2024-09-12
bump timeout for config sync autopilot tests

sdowell:
It appears that new versions of autopilot (currently only in rapid channel) scale up slower, and as a result the tests are occasionally timing out. The average runtime is very close to the current 2h timeout, causing it to sometimes pass and sometimes time out.

A 4h timeout matches the current timeout set on the prow job config.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
@@ -39,7 +39,7 @@ prow_ignored:
       <<: *config-sync-base-job-container
       env:
       - name: GKE_E2E_TIMEOUT
-        value: 2h
+        value: 4h
       - name: GCP_PROJECT
         value: kpt-config-sync-ci-release
       - name: GCP_NETWORK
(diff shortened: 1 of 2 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics-release.yaml
```diff
@@ -39,7 +39,7 @@ prow_ignored:
       <<: *config-sync-base-job-container
       env:
       - name: GKE_E2E_TIMEOUT
-        value: 2h
+        value: 4h
```
nan-yu, line 42: This is the timeout for standard cluster CIs. I thought we only need to bump the timeout for autopilot ones.

sdowell: Yeah it's only strictly necessary for autopilot at the moment. I figure these may as well be aligned with the timeout for the prow job, which is set to 4h for all jobs. I can make them different if you prefer

nan-yu: ACK. We can keep it the same as the prow job.

nan-yu (approved):
/lgtm

PR #2379, opened 2024-09-18
Remove gotest from galog's presubmits

drewhli:
It seems the test aren't run with root permissions, causing the persistent presubmit failures. Since these tests are already automatically run as presubmits beforehand, it should be fine to remove these.

/cc @dorileo @ChaitanyaKulkarni28


drewhli:
Instead of removing the entire testing, we're only going to skip the tests that are failing.

PR #2382, opened 2024-09-24
Revert "create new clusters for autopilot-rapid tests"

sdowell:
This reverts commit 8698884a1f6b4b89e05e439936dadd67e35350db.

Reason for revert: RCA is complete for these clusters, so the tests can switch back to the original cluster names.


nan-yu (approved):
/lgtm

PR #2384, opened 2024-09-27
Seperating cit-periodics to be per os

bkatyl:
- setting parallel_count=5 per OS
- adding rhel-sap
- added missing oses
- cit-periodics-performance --> -parallel_count=2


a-crate:
Note that this will result in shorter bursts of higher capacity, going from 20 VMs at once to 55 VMs at once on inital trigger (and we can't stagger them without staggering the submit timing of commits adding the jobs, IIUC). Should be fine, this still two orders of magnitude lower than the per-project CPU quota limit.

The timeout can be reduced for each of these jobs, I'd suggest, and this is a conservative suggestion to avoiding accidentally timing out jobs too fast, `2h x number of OSs / parallel_count`.
Rounding up, that's 6h for windows, 4h for ubuntu-os-pro, 3h for ubuntu, rhel-sap, rocky, 2h for debian, suse and rhel, 1h for opensuse, centos, and alma

But that's not super necessary, since there's no one big job that needs to get killed if it's taking too long anymore.
/lgtm
/hold for author

PR #2385, opened 2024-09-27
Bind sinker to the same GSA as prow-controller-manager.

cjwagner:
I noticed that sinker can't connect to some build clusters, presumably due to switching auth for those clusters to rely on workload identity since sinker is missing that.

I've already granted `roles/iam.workloadIdentityUser` to `oss-prow.svc.id.goog[default/sinker]` so this should take affect after this PR merges and we restart sinker.

/assign @michelle192837


michelle192837 (approved):
Thanks Cole!

PR #2391, opened 2024-10-01
Mark optional jobs in gke-networking-api.

yswe:
The `regenerate-crds` job is only triggered on PRs created by dependabot. This prevents automatic merges for other PRs in the GoogleCloudPlatform/gke-networking-api repository. Marking the job as optional will allow automatic merges to proceed.


cjwagner (approved):
LGTM

PR #2413, opened 2024-12-11
Lower accelerator specific cit-peridiocs to once per day

a-crate:
We have limited capacity, we need to be more economical with how we use it. This is enough for a regression signal.

This should help our image builds find capacity more easily.

/cc @zmarano @lpleahy


a-crate:
Update to use the cron field so we can force it run at 8PM.

PR #2437, opened 2025-02-19
Add external secret from build cluster for openshift-eng

sjswerdlow:
CL Auto generated by create-build-cluster.sh.


nathanperkins:
Confirmed I see the secret in the oss-prow project.

PR #2439, opened 2025-02-20
Modernize create-build-cluster.sh to use gke-auth-plugin kubeconfigs rather than gencred+KES.

cjwagner:
This doesn't automatically create a CL anymore, but we can iterate on that later if desired. Overall, this onboarding process is simpler.

cc @AlexBasinov 
/assign @simony-gke


cjwagner:
I split this into two commits so it is easier to see the diff of the changes that I actually made to the script after copying it here.

PR #2448, opened 2025-03-07
Stop LGTM acting as approve in Kubeflow

thesuperzapper:
Currently in the Kubeflow org, we have issues with people LGTMing PRs that are in root approvers, causing the PR to merge immediately.

So we can separate the idea of LGTM and APPROVE, this PR makes it so that an LGTM does not also APPROVE.

It also makes it so that we don't have to explicitly list out our repos anymore, an just applies the configs to the whole `kubeflow` org.

### DIFF
--- a/prow/oss/plugins.yaml
+++ b/prow/oss/plugins.yaml
@@ -47,24 +47,11 @@ approve:
 - repos:
   - GoogleCloudPlatform/k8s-config-connector
   - GoogleContainerTools
-  - kubeflow/arena
-  - kubeflow/blog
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@andreyvelich can confirm we want this change

andreyvelich (commented):
Thanks @thesuperzapper!
/lgtm
/cc @johnugeorge @terrytangyuan @franciscojavierarceo @juliusvonkohout

michelle192837 (approved):
Sounds good! LGTM

PR #2449, opened 2025-03-13
kpt-config-sync: more nodes for stress tests

karlkfi:
The stress tests need enough nodes to schedule 1,000 pods. The pods are small, but each node has a 110 pod limit. Control plane initial scale is dependent on number of vCPUs, so fewer cores per node should work with more nodes.

Since the stress tests extend the non-stress test template, I've hard coded the values in the stress config, so we can remove them from the non-stress test template later, and fall back to dynamic defaults in the test code.


sdowell (approved):
/lgtm

PR #2456, opened 2025-03-26
kpt-config-sync: increase stress test memory

karlkfi:
The stress tests don't have a memory limit, but they're often getting OOMKilled by the kernel during some of the more memory intensive tests. So increase the memory request to 20GiB to avoid node conjestion.

### DIFF
--- a/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
+++ b/prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
@@ -399,6 +399,10 @@ periodics:
       - 'GKE_MACHINE_TYPE=n2-standard-4'
       - 'GKE_DISK_SIZE=25Gb'
       - 'E2E_ARGS=--stress'
+      resources:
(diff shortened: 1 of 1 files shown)


prow/prowjobs/GoogleContainerTools/kpt-config-sync/kpt-config-sync-periodics.yaml
```diff
@@ -399,6 +399,9 @@ periodics:
       - 'GKE_MACHINE_TYPE=n2-standard-4'
       - 'GKE_DISK_SIZE=25Gb'
       - 'E2E_ARGS=--stress'
+      resources:
+        requests:
+          memory: "20Gi" # Stress tests watch many large objects
```
sdowell, line 404: Suggest setting a cpu request as well.

Given the shape of the nodes I'd suggest something like 6cpu / 24Gi mem

karlkfi: done

sdowell (approved):
/lgtm

PR #2457, opened 2025-03-26
kpt-config-sync: more stress test nodes

karlkfi:
TestStressManyDeployments needs at least 13x 4cpu  nodes to have enough cpu for 1k pods requesting 50m each.


sdowell (approved):
/lgtm
/approve

PR #2458, opened 2025-04-07
Add a new build cluster to the OSS Prow cluster’s configuration.

AlexBasinov:
This PR adds a new build cluster to the  OSS Prow cluster’s configuration.


cjwagner (commented):
This configures default settings for the new build cluster, but the build cluster also needs to be registered by adding an entry to kubeconfigs.yaml. The script output should include some content for you to copy to that file.

cjwagner (approved):

PR #2461, opened 2025-04-09
Remove all references to the old build cluster

AlexBasinov:
In https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2458, we registered a new build cluster to run presubmit jobs. However, these jobs are now failing with the following error:

"Terminal error: nonretryable error: no build client found for cluster
    alias "build-gcp-oracle-team". If this was previously working, then it''s likely
    that the credentials expired, there was a CA cert expiry, or we cannot reach your
    cluster API for some other reason."

Failed prow job for reference: [Prow job link ](https://oss.gprow.dev/prowjob?prowjob=c4b45b30-4535-4b21-a924-dddba384ab01)

Based on an internal Q&A thread, filing this PR to remove the old cluster context from the legacy kubeconfig secret to address the issue.

### DIFF
--- a/prow/oss/cluster/crier.yaml
+++ b/prow/oss/cluster/crier.yaml
@@ -37,7 +37,7 @@ spec:
         env:
         # Use KUBECONFIG envvar rather than --kubeconfig flag in order to provide multiple configs to merge.
         - name: KUBECONFIG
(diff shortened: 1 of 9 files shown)


cjwagner:
/hold
I don't think this will address the problem. The old build cluster had a different name so there was no collision to resolve.

This looks like a permissions error, I shared more over chat.

cjwagner:
We will want to clean up the references to the old build cluster once it is no longer used so this is worth keeping open.
We'll need to remove references in more than just prow-controller-manager though. https://github.com/search?q=repo%3AGoogleCloudPlatform%2Foss-test-infra%20build-bms-oracle-team&type=code

cjwagner (approved):

AlexBasinov:
> We will want to clean up the references to the old build cluster once it is no longer used so this is worth keeping open. We'll need to remove references in more than just prow-controller-manager though. https://github.com/search?q=repo%3AGoogleCloudPlatform%2Foss-test-infra%20build-bms-oracle-team&type=code

I’ve addressed this. Is there anything else needed before this can be merged?
Thanks!

cjwagner:
No nothing else is needed. We just need to remove the hold.
/unhold

PR #2515, opened 2025-07-30
set always_run to true and remove skip_report for single-instance-on-gcp test

AlexBasinov:
Activate the single-instance on gcp test, which was previously disabled (`always_run: false`) while still under development. Now that it is now ready, they will run unconditionally for all PRs.

### DIFF
--- a/prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
+++ b/prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
@@ -3,8 +3,7 @@ presubmits:
 
   - name: oracle-toolkit-install-single-instance-on-bms
     cluster: build-gcp-oracle-team
-    always_run: false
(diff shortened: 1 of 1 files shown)


AlexBasinov:
@mfielding I flipped `always_run` to false in the 1f97ab35d8d9498b6e145cb5af931c3b0470b708 commit because it cannot be used together with `skip_if_only_changed`.  They are mutually exclusive in Prow. 

I had previously filed a separate https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2512 PR to enable the GCP tests by setting always_run: true, but that PR is now redundant. I think it's better to trigger presubmit tests conditionally using skip_if_only_changed. So that separate PR is no longer needed, and this one covers both changes.

AlexBasinov:
> Will these tests continue to function as presubmit tests?
> 
> The issue with postsubmit continuous tests is: what happens if they fail? Are we hoping someone will read an e-mail?

Yes, according to https://docs.prow.k8s.io/docs/jobs/#triggering-jobs the tests will act as presubmit tests.

mfielding (approved):

PR #2535, opened 2025-09-30
Set required env vars for gcenode

Camila-B:
There are currently no tests that run as part of https://testgrid.corp.google.com/googleoss-kpt-config-sync-main#standard-regular-gcenode. This change adds the required repository configs to enable the gcenode tests


sdowell (approved):
Good catch! Thank you for fixing

/lgtm

PR #2591, opened 2026-04-15
Add StickyLgtmTeam for gke-mcp repo

bradhoekstra:
Add trusted_team_for_sticky_lgtm to make LGTM behavior more like google3.


GinnyJI:
Does this make a big followup commit (like a logic change) being able to submit without review as well?

bradhoekstra:
> Does this make a big followup commit (like a logic change) being able to submit without review as well?

Yes, similar to internal this implies some trust in the author after granting LGTM. This is restricted to our Contributors team, which is constricted to GCP org members.

cjwagner (approved):

PR #2593, opened 2026-04-27
Update plugins.yaml

bradhoekstra:
Disable blunderbuss plugin for gke-mcp repo.

We don't like the 2 assigned reviewers, and blunderbuss is not configurable on the repo level. We are going to use Github's built in auto assign feature instead.

### DIFF
--- a/prow/oss/plugins.yaml
+++ b/prow/oss/plugins.yaml
@@ -342,7 +342,6 @@ plugins:
   GoogleCloudPlatform/gke-mcp:
     plugins:
     - approve
-    - blunderbuss
     - lgtm
     - owners-label
     - size


bradhoekstra:
/hold

I want to test what happens if we only have approvers first (this didn't work, still get 2 people assigned).

cjwagner (approved):
FYI blunderbuss is configurable if you want to tweak it's behavior instead. Totally up to whatever you prefer.

```
blunderbuss:
    # ExcludeApprovers controls whether approvers are considered to be
    # reviewers. By default, approvers are considered as reviewers if
    # insufficient reviewers are available. If ExcludeApprovers is true,
    # approvers will never be considered as reviewers.
    exclude_approvers: true
    # MaxReviewerCount is the maximum number of reviewers to request
    # reviews from. Defaults to 0 meaning no limit.
    max_request_count: 3
    # ReviewerCount is the minimum number of reviewers to request
    # reviews from. Defaults to requesting reviews from 2 reviewers
    request_count: 2
    # UseStatusAvailability controls whether blunderbuss will consider GitHub's
    # status availability when requesting reviews for users. This will use at one
    # additional token per successful reviewer (and potentially more depending on
    # how many busy reviewers it had to pass over).
    use_status_availability: true
```

(This comes from https://oss.gprow.dev/plugins by the way. The page is very slow to initially load due to listing lots of GH repos, but it eventually works.)

bradhoekstra:
> FYI blunderbuss is configurable if you want to tweak it's behavior instead. Totally up to whatever you prefer.

I saw that, but from what I can tell it isn't configurable on a per-repo basis (other plugin configs are arrays, that accept an array of repos as keys). I don't want to change the config for everyone.

PR #2603, opened 2026-06-09
Add a new presubmit job to run Python unit tests

AlexBasinov:
Add a new presubmit job to run Python unit tests for the inventory plugin.

The presubmit job is only triggers if files in the inventory_plugins directory are changed.

### DIFF
--- a/prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
+++ b/prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
@@ -176,12 +176,12 @@ presubmits:
     spec:
       serviceAccountName: prowjob-default-sa
       containers:
-      - image: mirror.gcr.io/library/python:3.14.5-slim
+      - image: quay.io/ansible/ansible-runner:stable-2.9-latest
         command:
          - /bin/sh
          - -c
          - |
-           pip install ansible-core pytest PyYAML
+           pip install pytest
            pytest inventory_plugins/
         resources:
           requests:
--- a/prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
+++ b/prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
@@ -164,3 +164,26 @@ presubmits:
           requests:
             memory: "2.0Gi"
             cpu: "3.0"
+  - name: oracle-toolkit-python-unit-tests
+    cluster: build-gcp-oracle-team
+    # Only triggers if files in the inventory_plugins directory are changed
+    run_if_changed: "^inventory_plugins/"
+    max_concurrency: 3
+    decorate: true
+    decoration_config:
+      timeout: 4h
+      grace_period: 5m
+    spec:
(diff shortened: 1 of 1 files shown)


prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
```diff
@@ -164,3 +164,26 @@ presubmits:
...
+      grace_period: 5m
+    spec:
+      serviceAccountName: prowjob-default-sa
+      containers:
+      - image: mirror.gcr.io/library/python:3.14.5-slim
+        command:
+         - /bin/sh
+         - -c
+         - |
+           pip install ansible-core pytest PyYAML
```
mfielding: Hi @AlexBasinov,

Thanks for the PR! I noticed the new job `oracle-toolkit-python-unit-tests` uses `python:3.14-slim` and installs the latest `ansible-core` and `pytest` dynamically at runtime.

I understand we want to avoid stale dependencies and automatically inherit upstream security updates (aligned with organizational mandates). However, doing this via unpinned `pip install` at runtime introduces some risks:

1.  **Version Mismatch with Integration Tests:** The existing job (`oracle-toolkit-install-single-instance-on-bms`) is pinned to Ansible 2.9 via `quay.io/ansible/ansible-runner:stable-2.9-latest`. Since inventory plugins are tightly coupled with Ansible internals, testing them on the latest `ansible-core` while they run on 2.9 in integration/production may lead to false positives (tests pass but fail in prod) or false negatives (tests fail on deprecations that don't affect 2.9).
2.  **Runtime Breakage Risk:** If an upstream dependency releases a breaking change, the presubmit job will suddenly start failing, blocking unrelated PRs.
3.  **Supply Chain Risk:** Pulling unpinned packages at runtime increases vulnerability to dependency confusion or compromised package versions.

**Questions/Suggestions:**
*   **Are we upgrading the target Ansible version?** If so, we should probably update the integration test job to match.
*   **If we must support 2.9,** should we run these unit tests inside the `ansible-runner:stable-2.9-latest` image instead (only installing `pytest`)?
*   **Alternative for Auto-Updates:** To balance security updates with stability, would you be open to pinning the versions in a `requirements.txt` and using an automated dependency updater (like Renovate/Dependabot) to manage bumps? This ensures security patches are applied quickly but are verified by CI before merging.

What are your thoughts on how we should align these environments?

mfielding (approved):

prow/prowjobs/google/oracle-toolkit/oracle-toolkit-presubmit.yaml
AlexBasinov: I've verified that inventory plugins's unit tests work correctly on python:3.14-slim image before filing this PR, but I agree -  this new presubmit job should use the same image used for other presubmit tests.

I've pushed a new commit to set the image to ansible-runner:stable-2.9-latest to match the existing presubmit tests.

As for pinning dependency versions in a requirements.txt, I think this is unnecessary because both ansible-core and PyYAML dependencies come with the ansible-runner:stable-2.9-latest image by default.

PR #2641, opened 2026-07-22
update cluster name to trigger cluster refresh for regional cluster

Sneha-at:
Current build cluster was updated from zonal to regional in https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2634 but the latest prow job is still using the zonal cluster. This PR updates the cluster name so that Prow picks regional cluster


nathanperkins (approved):

Sneha-at:
The regional cluster is now working so no need to update the name

PR #2648, opened 2026-08-05
configs(kube-agents): replace devops-bench eval image with kubekins-e2e base image to run evals via library dependency

lapis2002:
### Summary
Replace the `devops-bench-harness` container image in `pull-kube-agents-smoke-test` with the standard Prow base image (`kubekins-e2e`).

### Why
This change works together with [kube-agents#466](https://github.com/gke-labs/kube-agents/pull/466) to decouple `kube-agents` from `devops-bench`. By consuming `devops-bench` as a library dependency via `uv`, the presubmit job no longer depends on a custom `devops-bench` Docker image.

### Testing
- Verified in a local container (`gcr.io/k8s-staging-test-infra/kubekins-e2e:latest-1.32`) that `uv`, OpenTofu, and `gcloud` are installed.

### DIFF
--- a/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
+++ b/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
@@ -11,13 +11,16 @@ presubmits:
     spec:
       serviceAccountName: prowjob-default-sa
       containers:
-      - image: us-central1-docker.pkg.dev/kube-agents-prow/kube-agents/devops-bench-harness:latest
+      - image: gcr.io/k8s-staging-test-infra/kubekins-e2e:latest-1.32
         command:
         - /bin/bash
         - -c
         - |
           set -euo pipefail
-          apt-get update && apt-get install -y --no-install-recommends make gettext-base golang-go git openssl ca-certificates
+          apt-get update && apt-get install -y --no-install-recommends make gettext-base golang-go git openssl ca-certificates curl
+          curl -LsSf https://astral.sh/uv/install.sh | sh
+          export PATH="$HOME/.local/bin:$PATH"
+          curl --proto '=https' --tlsv1.2 -fsSL https://get.opentofu.org/install-opentofu.sh -o install-opentofu.sh && chmod +x install-opentofu.sh && ./install-opentofu.sh --install-method standalone
           export GEMINI_API_KEY="$(cat /etc/gemini-secret/GEMINI_API_KEY)"
           trap './hack/ci-teardown.sh' EXIT
           ./hack/ci-connection-test.sh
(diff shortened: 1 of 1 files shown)


prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
```diff
@@ -11,13 +11,16 @@ presubmits:
     spec:
       serviceAccountName: prowjob-default-sa
       containers:
-      - image: us-central1-docker.pkg.dev/kube-agents-prow/kube-agents/devops-bench-harness:latest
+      - image: gcr.io/k8s-staging-test-infra/kubekins-e2e:latest-1.32
```
jayantid, line 14: I think this is missing `gke-gcloud-auth-plugin`. kubekins-e2e is `FROM bootstrap`, and bootstrap installs gcloud with `gcloud components install alpha beta kubectl` — no auth plugin, and
  neither Dockerfile adds one. The image you're replacing installed `google-cloud-cli-gke-gcloud-auth-plugin` explicitly.

  Since k8s 1.26 that binary is what the kubeconfig's `exec` provider calls, so `get-credentials` will succeed and the first kubectl after it dies with `executable gke-gcloud-auth-plugin not
  found`. That hits `ci-connection-test.sh` and then `ci-deploy.sh:88`.

  Worth flagging that the green runs on kube-agents#582 don't cover this — Prow takes presubmit definitions from deployed config, not from a PR branch, so those ran on
  `devops-bench-harness:latest` with all the tooling baked in. This image path hasn't actually executed yet.

  ```bash
  gcloud components install gke-gcloud-auth-plugin --quiet

jayantid (changes requested):

prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
lapis2002: updated.

jayantid (approved):
Hey Hang, 

Since the first real execution of this will be after merge, I would de-risk by running the setup lines against the image directly, which would have caught the missing plugin:

  ```bash
  docker run --rm -it gcr.io/k8s-staging-test-infra/kubekins-e2e:latest-1.32 bash -c '
    set -euo pipefail
    apt-get update && apt-get install -y --no-install-recommends gettext-base
    gcloud components install gke-gcloud-auth-plugin --quiet
    curl -LsSf https://astral.sh/uv/install.sh | sh
    export PATH="$HOME/.local/bin:$PATH"
    curl --proto "=https" --tlsv1.2 -fsSL https://get.opentofu.org/install-opentofu.sh -o /tmp/i.sh
    chmod +x /tmp/i.sh && /tmp/i.sh --install-method standalone
    command -v gke-gcloud-auth-plugin uv tofu && tofu version && uv --version'
  ```

Would you be able to verify this? Merging is low risk but I would verify in the background.

Thanks

PR #2658, opened 2026-08-21
fix(kube-agents): bound heartbeat retries to 8 and expand markdown doc skip filter in smoke test

lapis2002:
## Summary

1. Bounds background `boskosctl heartbeat` retries to 8 and triggers fail-fast termination (`kill -TERM $$`) if the Boskos lease heartbeat fails in `pull-kube-agents-smoke-test`.
2. Fixes `skip_if_only_changed` to match any `.md` markdown file across the repository (e.g. `agents/contributor/AGENTS.md`) rather than only root-level `.md` files.

## Context & Motivation

- **Boskos Heartbeat Bounds:** The Boskos reaper in `gke-internal/test-infra` reclaims stale leases after 5 minutes of silence (`--expire=5m`). The default `boskosctl heartbeat` with 10 retries at 30s intervals (5m30s) runs detached in the background; setting `--retries=8` bounds retry attempts to ~4m30s (safely inside the 5m reaper window) and `|| kill -TERM $$` sends `SIGTERM` to trigger `cleanup()` immediately upon lease loss.
- **Skip Filter Expansion:** `skip_if_only_changed` previously used `[^/]+\.md$`, which only matched root-level markdown files and failed to skip documentation changes in subdirectories (such as `agents/contributor/AGENTS.md` in gke-labs/kube-agents#876), causing unnecessary ~43m smoke tests on doc-only PRs. Changing to `\.md$` ensures all markdown files anywhere in the repo tree skip the smoke test.

## Changes

- `prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml`:
  - Update `skip_if_only_changed` regex to `^(docs/|\.github/|examples/)|\.md$|^(LICENSE|OWNERS|OWNERS_ALIASES)$`.
  - Update heartbeat invocation to `( "${BOSKOSCTL[@]}" heartbeat --resource "${RESOURCE}" --period=30s --retries=8 || kill -TERM $$ ) &`.

## Testing

- `go test -v -count=1 ./prow/tests/` (PASS).
- Evaluated `skip_if_only_changed` Go regexp engine across changed file paths:
  - Validated doc-only file sets (e.g. gke-labs/kube-agents#876 containing `agents/contributor/AGENTS.md`, `AGENTS.md`, `docs/README.md`) evaluate to `Skip: true`.
  - Validated code paths (`hack/ci-deploy.sh`, `k8s-operator/...`, `charts/...`, `admin_console/...`) evaluate to `Skip: false`.

### DIFF
--- a/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
+++ b/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
@@ -11,7 +11,7 @@ presubmits:
     cluster: build-kube-agents
     always_run: false
     skip_report: false
-    skip_if_only_changed: '^(docs/|\.github/|examples/|[^/]+\.md$|LICENSE$|OWNERS$|OWNERS_ALIASES$)'
+    skip_if_only_changed: '^(docs/|\.github/|examples/)|\.md$|^(LICENSE|OWNERS|OWNERS_ALIASES)$'
     # Reported on every PR but not merge-blocking: the job provisions a GKE
     # cluster and runs a full eval, so it is currently too slow to gate Tide on.
     # Drop this once the runtime is down.
@@ -76,7 +76,7 @@ presubmits:
           }
           trap cleanup EXIT INT TERM
 
-          "${BOSKOSCTL[@]}" heartbeat --resource "${RESOURCE}" --period=30s &
+          ( "${BOSKOSCTL[@]}" heartbeat --resource "${RESOURCE}" --period=30s --retries=8 || kill -TERM $$ ) &
           HEARTBEAT_PID=$!
 
           # Clean up any leftover state if a previous job was hard-killed
--- a/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
+++ b/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
@@ -11,7 +11,9 @@ presubmits:
     cluster: build-kube-agents
     always_run: false
     skip_report: false
-    skip_if_only_changed: '^(docs/|\.github/|examples/)|\.md$|^(LICENSE|OWNERS|OWNERS_ALIASES)$'
+    # .md is root-anchored: agents/**/*.md is prompt content shipped in the
+    # image (deploy/docker/Dockerfile), so those PRs must still run the eval.
+    skip_if_only_changed: '^(docs/|\.github/|examples/)|^[^/]+\.md$|^(LICENSE|OWNERS|OWNERS_ALIASES)$'
     # Reported on every PR but not merge-blocking: the job provisions a GKE
     # cluster and runs a full eval, so it is currently too slow to gate Tide on.
     # Drop this once the runtime is down.
@@ -66,6 +68,7 @@ presubmits:
             local rc=$?
             trap - EXIT INT TERM
             kill "${HEARTBEAT_PID:-}" 2>/dev/null || true
+            kill "${STEP_PID:-}" 2>/dev/null || true
             echo "=== [$(date -u +'%Y-%m-%dT%H:%M:%SZ')] Running CI Teardown ==="
             ./hack/ci-teardown.sh || true
             if [ -n "${LEASED_PROJECT:-}" ] && [ "${LEASED_PROJECT}" != "null" ]; then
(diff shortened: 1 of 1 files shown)


prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
```diff
@@ -11,7 +11,7 @@ presubmits:
     cluster: build-kube-agents
     always_run: false
     skip_report: false
-    skip_if_only_changed: '^(docs/|\.github/|examples/|[^/]+\.md$|LICENSE$|OWNERS$|OWNERS_ALIASES$)'
+    skip_if_only_changed: '^(docs/|\.github/|examples/)|\.md$|^(LICENSE|OWNERS|OWNERS_ALIASES)$'
```
jayantid: `\.md$` matches every markdown file in the repo, but in kube-agents much of the markdown under `agents/platform/` is runtime prompt content, not documentation. `deploy/docker/Dockerfile:1233-1245` COPYs `agents/platform/AGENTS.md`, `SOUL.md`, `CAPABILITIES.md`, `docs/glossary.md`, and the whole `skills/`, `governance/`, `cron/`, and `hindsight/` trees into the image.

A PR that only edits `agents/platform/skills/*/SKILL.md` or a governance SOP changes deployed agent behavior. Under this filter every file in that PR matches `\.md$`, so the smoke test is skipped. The eval is the only end-to-end check of prompt content, and it now silently never runs for exactly the class of change it exists to catch. Prompt-tuning PRs are routine in this repo, so this fires deterministically, not rarely. The motivating file in the description, `agents/contributor/AGENTS.md`, sits next to `agents/platform/AGENTS.md` — which ships in the image.

Go RE2 has no negative lookahead, so "any .md except under agents/" can't be expressed directly. Keep the skip scoped to root-level markdown and true doc trees, and add specific subtrees only after confirming the image doesn't ship them:

```yaml
skip_if_only_changed: '^(docs/|\.github/|examples/)|^[^/]+\.md$|^(LICENSE|OWNERS|OWNERS_ALIASES)$'
```

prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
```diff
@@ -76,7 +76,7 @@ presubmits:
           }
           trap cleanup EXIT INT TERM
 
-          "${BOSKOSCTL[@]}" heartbeat --resource "${RESOURCE}" --period=30s &
+          ( "${BOSKOSCTL[@]}" heartbeat --resource "${RESOURCE}" --period=30s --retries=8 || kill -TERM $$ ) &
```
jayantid: `kill -TERM $$` cannot interrupt the run. Bash defers a trapped signal until the current foreground command completes (man bash, SIGNALS), and after this line the script is always inside one of the `./hack/ci-*.sh` foreground commands.

Two failure modes, both reproduced with this script's exact trap/cleanup structure. (a) Lease lost mid-eval: the SIGTERM stays pending for the rest of the ~40m `ci-eval-pr.sh`, the job keeps using the project after the 5m reaper frees it, and a concurrent run (`max_concurrency: 2`) can lease and deploy into the same project — the race the description says this closes "immediately upon lease loss" stays open. (b) Lease lost during a step that exits 0, e.g. `ci-deploy.sh`: the deferred trap runs `cleanup` with `rc=$?` equal to 0, so the job exits green having skipped `ci-eval-pr.sh` entirely. That is a passing presubmit with no eval run.

Signal the process group instead. The foreground child dies, the trap fires promptly, and `cleanup` sees rc=143:

```bash
( "${BOSKOSCTL[@]}" heartbeat --resource "${RESOURCE}" --period=30s --retries=8 || kill -TERM 0 ) &
```

Verified in an isolated process group: `kill -TERM 0` interrupts the foreground command immediately and the job exits 143, while `kill -TERM $$` leaves it running and can exit 0.

prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
lapis2002: Applied your suggestion.

There would be trade off that nested doc-only READMEs (bench/, k8s-operator/, charts/kube-agents/) no longer skip. That's the safe direction given RE2 has no negative lookahead

prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
lapis2002: Confirmed and fixed, though not quite as suggested; reasoning below.
 
**The fix.** Bash can't interrupt a *foreground* command, but it can interrupt `wait`. So each step is backgrounded and waited on, and `kill -TERM $$` stays:
```
      run_step() { "$@" & STEP_PID=$!; wait "${STEP_PID}"; }
```
**Why not `kill -TERM 0`.** Just as prompt, but only if the job has its own process group. If prow's entrypoint starts it without `Setpgid`, that group also holds the entrypoint — which would mark the job aborted and SIGKILL the script at `grace_period` (5m), cutting `ci-teardown.sh` and the Boskos release short.

Also: `cleanup()` now kills `STEP_PID` so a step can't outlive the trap, and backgrounded steps get stdin from `/dev/null` — checked that none of the four steps or the `teardown_*.sh` scripts they call read stdin.

jayantid (approved):

PR #2663, opened 2026-08-24
Update base_ref to v1.25 for kpt-config-sync release periodics

cowsking:
Update base_ref to v1.25 for kpt-config-sync release periodics as part of Config Sync 1.25.0 release.


tiffanny29631 (approved):
/lgtm

PR #2670, opened 2026-08-26
config(kube-agents): raise smoke-test max_concurrency to 6

lapis2002:
The kube-agents-evals-project Boskos pool has grown from three projects to 6. `max_concurrency` is the tap and the pool is the ceiling, so this raises the tap to match.

`kube-agents-evals-4`, `-5` and `-6` completed provisioning on 2026-08-26 and passed  every check in `scripts/verify_ci_pool_project.py` — APIs, Workload Identity, Artifact Registry, the four clusters and CMEK, the seven seeded fleet fixtures, the GitOps repo, and a KMS signature accepted by GitHub as App 4675512. Exit 0 on all six.

Also corrects a comment that still said the App is installed on three `*-infra` repos; it is six. That list, not this value, is what bounds where a run can write.

Context: gke-labs/kube-agents#903.

### DIFF
--- a/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
+++ b/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
@@ -2,14 +2,17 @@ presubmits:
   gke-labs/kube-agents:
   - name: pull-kube-agents-smoke-test
     # Boskos dynamically leases an evaluation project per run from the
-    # kube-agents-evals-project pool, which holds three fully provisioned
-    # projects (kube-agents-evals, -2, -3 — the third completed 2026-08-24).
-    # max_concurrency matches the pool size, so the pool runs saturated: a
-    # project stuck dirty degrades throughput to 2 rather than queueing work.
-    max_concurrency: 3
+    # kube-agents-evals-project pool, which holds six fully provisioned
+    # projects (kube-agents-evals through -6 — the last three completed
+    # 2026-08-26). max_concurrency matches the pool size, so the pool runs
+    # saturated: a project stuck dirty degrades throughput to 5 rather than
+    # queueing work.
+    max_concurrency: 6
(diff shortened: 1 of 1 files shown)


jayantid (approved):

prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
```diff
@@ -2,14 +2,17 @@ presubmits:
...
-    # projects (kube-agents-evals, -2, -3 — the third completed 2026-08-24).
-    # max_concurrency matches the pool size, so the pool runs saturated: a
-    # project stuck dirty degrades throughput to 2 rather than queueing work.
-    max_concurrency: 3
+    # kube-agents-evals-project pool, which holds six fully provisioned
+    # projects (kube-agents-evals through -6 — the last three completed
+    # 2026-08-26). max_concurrency matches the pool size, so the pool runs
+    # saturated: a project stuck dirty degrades throughput to 5 rather than
+    # queueing work.
+    max_concurrency: 6
```
jayantid: 🟡 **Not merge-blocking if the Boskos entry has landed — hold until that's confirmed.**

The PR says the pool "has grown from three projects to 6", but the evidence it cites stops one step short. `verify_ci_pool_project.py` is the pre-registration gate: its success line reads "safe to register in Boskos", and its docstring says to run it before the Boskos entry lands in gke-internal/test-infra. Exit 0 on -4/-5/-6 proves they are ready to register, not registered.

If the entry hasn't landed, this change is worse than today. Runs 4–6 don't queue the way `max_concurrency: 3` queues them — they start, sit in `boskosctl acquire --timeout=10m` (line 121), and fail red after 10 minutes, since runs hold leases for ~43–78m. That is the "raise before register" case the comment on line 15 warns about.

Fix: link the gke-internal/test-infra change that registers the three projects, then merge.

PR #2676, opened 2026-08-31
fix(kube-agents): raise the smoke deadline to 360m; 3 repetitions stay

jayantid:
Reversal of this PR's first commit after review discussion: **three repetitions are the presubmit's flake tolerance** — the collapse rule (a case reds only by failing 3 of 3) is what keeps a single wobble from blocking a pull request, so they stay.

The problem is the deadline, not the repetitions: the measured 3-repetition suite at 16 cases spans 170–250+ minutes (good-day green at 170m — gke-labs/kube-agents#1053's run; two weekend runs killed at the 240m wire with 10+ tasks passing: builds 2094208…250304 and 2094333…343488). Negative margin on any slow day, most of the spread being one audit case's per-repetition variance. 360m restores 1.4–2× headroom over worst observed.

Second commit reverts the EVAL_REPETITIONS=1 export from the first; net diff vs master is the timeout line only.

### DIFF
--- a/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
+++ b/prow/prowjobs/gke-labs/kube-agents/kube-agents-presubmits.yaml
@@ -71,7 +71,7 @@ presubmits:
       # roughly a third more expensive than designed and 1.4x gets thin.
       # Revisit in either direction once several runs at three repetitions
       # exist.
-      timeout: 240m
+      timeout: 360m
       grace_period: 5m
     spec:
       serviceAccountName: prowjob-default-sa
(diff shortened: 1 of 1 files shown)


jayantid:
Weekend soak data hardening the case for this change: since gke-labs/kube-agents#1045 grew the suite to 16 cases, **no full 3-repetition run has finished green** — both runs that survived startup hit the 240m deadline outright (builds 2094208…250304 at ~241m with 10 tasks complete, and 2094333…343488 at ~242m). Pre-#1045 greens ran 167–201m, so the margin at 3 reps is gone entirely. At 1 repetition the same suite fits in ~90 min with ~150m of headroom. The interleaved early-death failures the same morning were a separate, now-diagnosed connectivity-check flake (fix in flight on the kube-agents side) — not a reason to hold this.

stalhaali (approved):

Issue #2687, opened 2026-09-03
tide: merge throughput collapses to one-per-job-duration for repos with multi-hour required presubmits; batching never engages (gke-labs/kube-agents)


jayantid:
> **EDIT 2026-09-03 (rewrite for accuracy):** the original body claimed tide batching never engages, based on an empty GCS path — that path was wrong and the claim is withdrawn (see thread). This rewrite is the current, unified diagnosis; the comment history preserves the corrections.

## Summary

`gke-labs/kube-agents` made its eval presubmit (`pull-kube-agents-smoke-test`, 1.5–3.5h wall clock) merge-blocking on 2026-09-02 (#2677). Since then, merge throughput has collapsed — at one point 13 PRs sat simultaneously `lgtm`+`approved` while merges landed hours apart, and maintainers force-merged three PRs in ~24h to unblock incident fixes. Tide appears to be behaving as designed; the design assumptions just interact badly with a multi-hour required job. We're asking for configuration guidance.

## Diagnosis: both of tide's validation paths break at this job length

**Serial path.** Tide merges a PR only when its required job passed against current `main`. Every merge moves `main`, invalidating every other pool PR's green result; each revalidation costs a full job-duration. Ceiling: ~1 merge per 1.5–3.5h regardless of how many PRs are ready.

**Batch path.** Tide does batch (confirmed — see the observed timeline below), which should amortize one validation across N PRs. But a batch result is discarded the moment ANY member leaves the pool — including when a member's own *individual* presubmit flakes while the batch is in flight. With a 2–3h in-flight window, there is ample time for a member to flake out, and the whole batch — **even a green one** — is thrown away.

Observed timeline (kube-agents#1191, an incident fix, approved 01:53Z, merged 09:53Z after ~8h):
- 01:55Z: tide batches #1177+#1191. **04:47Z: the batch PASSES.** It is discarded anyway — #1177's individual smoke had flaked at 04:26Z, dropping it from the pool.
- 07:14Z: tide re-batches #1191+#1152; still running 2.5h later.
- A maintainer broke the loop manually: `/hold` on #1191 so #1152 sat alone (a one-PR pool can't batch) and merged solo at 09:45Z, then released; #1191 was ultimately admin-merged.

Batch survival is the product of every member's stability over the batch's multi-hour flight. At last week's flake rates that product was ~zero; we have since eliminated the major flake classes repo-side, which revives both paths — but the structural fragilities remain.

## Questions for the owners

[...]

PR #2688, opened 2026-09-03
feat(kube-agents): cap tide batches at 2 and pin prioritize_existing_batches

jayantid:
Supersedes #2683 (which proposed disabling batching under a since-withdrawn theory — full correction history on #2687).

## What

Two tide settings for `gke-labs/kube-agents` only:
- `batch_size_limit: 2` — the default is **unlimited** ([kubernetes-sigs/prow pkg/config/tide.go](https://github.com/kubernetes-sigs/prow/blob/main/pkg/config/tide.go#L243): `0 => unlimited, -1 => disabled`). With a 1.5–3.5h merge-blocking presubmit, batch survival is the product of every member's stability over the flight window; on 2026-09-03 a **passing** two-PR batch was discarded because one member's individual retest flaked mid-flight, and the innocent PR sat mergeable ~8h (timeline in #2687). Pairs keep the 2× amortization while bounding one flake to one wasted partner. We'll propose raising to 3 if the repo's per-run green rate holds >95%.
- `prioritize_existing_batches: true` — upstream's default, pinned for legibility beside the cap: finish in-flight batches and reuse results rather than restarting when new PRs become eligible (a restart costs a full run).

## Why draft

Opened as draft pending the owners' read on #2687's questions (esp. whether green-batch salvage exists or is planned, which would change the calculus). If the answer is "tune batch size," this is the tuning; happy to adjust values to your guidance. #2683 will be closed in favor of this.

## Context

- #2687 — the throughput diagnosis (serial fresh-base retests + batch fragility at multi-hour job lengths).
- gke-labs/kube-agents#1202 — the repo-side plan this pairs with (flake elimination, wall-clock reduction).

### DIFF
--- a/prow/oss/config.yaml
+++ b/prow/oss/config.yaml
@@ -13,6 +13,26 @@
 # limitations under the License.
 
 tide:
+  batch_size_limit:
+    # Cap kube-agents tide batches at PAIRS. The repo's merge-blocking smoke
+    # presubmit runs 1.5-3.5h, and a batch is discarded whenever any member
+    # drops out of the pool mid-flight -- observed 2026-09-03: a PASSING
+    # two-PR batch was thrown away because one member's individual retest
+    # flaked while the batch ran (GoogleCloudPlatform/oss-test-infra#2687).
+    # Batch survival is the product of member stability over a multi-hour
+    # window, so wide batches almost never land; pairs keep the 2x
+    # amortization while bounding one member's flake to one wasted partner.
+    # Upstream semantics (kubernetes-sigs/prow pkg/config/tide.go):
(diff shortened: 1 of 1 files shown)


jayantid:
Closing after the source-level probe (findings on #2687): the batch cap of 2 harms the best case — an uncapped batch that survives merges all its members at once, and only one batch flies per subpool at a time, so the cap also caps amortization — while `prioritize_existing_batches` is provably irrelevant to the observed failures (it only biases membership selection for new batches and cannot salvage a batch whose member left the pool). With the repo-side flake eliminated, default batching is the right posture; the remaining structural asks (green-batch salvage, the batch-capture rule) live in #2687.

PR #2691, opened 2026-09-08
feat(kube-agents): add the post-kube-agents-eval-rc release-candidate postsubmit

stalhaali:
Adds `post-kube-agents-eval-rc`, an advisory eval of every `gke-labs/kube-agents` release candidate promoted to staging.

## Context

`gke-labs/kube-agents#1024` asks for every release candidate to run the full eval suite, non-gating, with results anyone can find. Two of its three pieces are merged in that repository already:

- **#1170** — `hack/resolve-rc-target.sh`, which answers "which commit is the candidate?" and refuses one whose images are not published.
- **#1230** — `hack/ci-eval-rc.sh`, the driver: resolve, `git checkout --detach`, deploy the candidate's **published** images (no rebuild), grade them, write a summary artifact.

Nothing invokes the driver. That script's header names a job in oss-test-infra as the caller and its own path as a contract. This is that job.

## Changed from the first revision: a postsubmit on the promotion tag, not a periodic

The original version of this PR was `ci-kube-agents-eval-rc`, a 02:00 UTC periodic. Review asked whether the eval could instead run when a candidate is promoted to staging, and it can — that is strictly better, so the job has been rewritten.

Prow fires postsubmits on tag pushes: `PushEvent.Branch()` strips `refs/tags/` as well as `refs/heads/`, and `branches` is matched against that value.

The periodic's stated reason for avoiding a tag trigger was a race — `rc-release-pipeline.yml` writes `rc_<ts>_<sha>` as its *first* step, while the images come from `docker-publish-ghcr.yml`, a separate push-to-main workflow with a serialising concurrency group, so an `rc_` tag can exist before its images do. That is real, and it is why the trigger here is on `staging_` and not on `rc_`. `nightly-pipeline.yml` pushes a staging tag only from step 5, which requires step 3 (the full E2E matrix) to have passed, and step 1 ran `scripts/release/verify_candidate_images.sh` before any of it. By the time a `staging_` tag exists the images have been verified present in GHCR and then installed and exercised. There is no window left for the resolver's image gate to fire.

What changes about coverage: the periodic measured the newest candidate nightly whether or not anything had been promoted, so most candidates were never measured and some were measured twice. This measures every promotion and nothing else. It therefore runs exactly as often as `nightly-pipeline.yml` promotes — today `workflow_dispatch` only, that file having no cron yet. Turning its schedule on is what makes this job nightly.

## What the job does

Leases a project from the `kube-agents-evals-project` pool exactly as `pull-kube-agents-smoke-test` does, then runs three steps instead of the presubmit's four:

```bash
run_step ./hack/ci-teardown.sh || true
run_step ./hack/ci-connection-test.sh
run_step ./hack/ci-eval-rc.sh || true
```

`ci-deploy.sh` and `ci-eval-pr.sh` are absent because the driver runs them itself, after the checkout, with `RC_COMMIT_SHA` set so `ci-deploy.sh` takes its published-image path. Invoking them here would rebuild images the promotion already published.

`RC_TAG` is pinned from `PULL_BASE_REF` rather than left unset, so the eval grades the commit this promotion tagged instead of whatever the newest candidate happens to be. `PULL_BASE_REF` and not `PULL_BASE_SHA`: the promotion tags are annotated, and the resolver peels `refs/tags/${RC_TAG}^{commit}` itself. If `PULL_BASE_REF` is somehow empty the job errors rather than falling back to the newest `rc_*`, which would publish a verdict under a commit nobody promoted.

The `branches` regex is the tag's exact shape, mirroring `STAGING_TAG_SHAPE_REGEX` in `scripts/release/common.sh`, rather than the bare `staging_` prefix that `staging-redeploy-*.yml` triggers on. The prefix is a trigger anyone can push by hand, and a `staging_hotfix` typed at a terminal would otherwise take a project out of a pool shared with the merge-blocking presubmit for hours.

## File placement

The periodic revision added `kube-agents-periodics.yaml`, which has since landed on `master` via `ci-kube-agents-pool-pressure`. This revision adds `kube-agents-postsubmits.yaml` instead, matching the type-split convention already used under `prow/prowjobs/GoogleContainerTools/kpt-config-sync/`. The earlier merge conflict is gone as a side effect.

## Testing

`go test ./prow/tests/` could not run on my machine — every freshly built Go binary is SIGKILLed there — so this needs the repo's own presubmit. What I did verify:

- **YAML parses**, and every field is checked against Prow's schema. `always_run` on a postsubmit has precedent here in `ESPv2-postsubmit-build` and `kpt-config-sync-publish-artifacts`.
- **No `extra_refs`, deliberately, and this was a bug in the first draft of the conversion.** A postsubmit already has a primary ref — the repo at the triggering tag — which decoration clones and makes the working directory (`DetermineWorkDir` falls through to `refs[0]`). Naming the same repo again in `extra_refs` does not add a checkout: both refs resolve to the same path, neither sets a `path_alias`, and `clonerefs`' `Options.Validate()` returns an error rather than a warning when colliding refs share an org and repo. That failure is in the init container, which `checkconfig` does not exercise, so the config would have validated green while every promotion died before the test container started.
- **`bash -n` on the embedded script** — clean. Nothing in CI checks this, and a syntax error surfaces only on a real promotion.
- **The resolver accepts a `staging_` tag.** `hack/resolve-rc-target.sh` applies no tag-family validation to an explicitly set `RC_TAG`; the `rc_*` glob is only the fallback for when it is empty.
- **Prow sets `CI=true`** for every decorated job (`pkg/pod-utils/downwardapi`). `release_fetch_tags` in the kube-agents repo is gated on it, and without the tag fetch the resolver finds no candidate.
- **`cluster: build-kube-agents` is not `test-infra-trusted`**, so `TestTrustedJobs` does not apply.
- No trailing whitespace, no tabs.

**Not live-tested.** This job cannot run until it is merged, and its first promotion will be the first execution of `hack/ci-eval-rc.sh` anywhere. I expect that run to surface something and will watch it.

## Known trade-off

`|| true` on the eval step means this job's status is uninformative: a RED candidate, a failed deploy, and a broken driver all report success — and as a postsubmit, that status is also a commit status on the promoted commit. That is deliberate and is where "non-gating" is expressed, per #1024's "nothing this lane ships may block a release this week". The driver preserves its real exit status specifically so the decision to ignore it lives visibly in this config rather than buried in a script. The verdict a reader wants is the `rc-eval-summary.md` artifact, which the driver writes on every path including the failure ones. `ci-connection-test.sh` is **not** wrapped, so a broken lease still reds the job.

---

Generated with the help of Claude (Opus 5).

### DIFF
--- a/prow/prowjobs/gke-labs/kube-agents/kube-agents-postsubmits.yaml
+++ b/prow/prowjobs/gke-labs/kube-agents/kube-agents-postsubmits.yaml
@@ -1,0 +1,365 @@
+postsubmits:
+  gke-labs/kube-agents:
+  - name: post-kube-agents-eval-rc
+    # The release-candidate eval (gke-labs/kube-agents#1024): checks the repo out
+    # at the commit a staging promotion just tagged, installs the images that
+    # were already PUBLISHED for it -- no rebuild -- and runs the eval catalog
+    # against them. Advisory: the verdict is reported and gates nothing, per that
+    # issue's "nothing this lane ships may block a release this week".
+    #
+    # This is the second of the "one job, two invocations" #1024 describes. The
+    # first is ci-kube-agents-eval-nightly, which measures main; this one measures
+    # the candidate. Almost everything below is copied from that job and from the
+    # pull-kube-agents-smoke-test presubmit -- same build cluster, same Boskos
+    # lease from the kube-agents-evals-project pool, same secrets, same heartbeat
+    # and cleanup handler. The differences are all in the step list and the three
+    # exports above it, and each one is commented where it appears.
+    #
+    # A POSTSUBMIT ON THE PROMOTION TAG, AND NOT A CRON. Prow fires postsubmits
+    # on tag pushes -- PushEvent.Branch strips refs/tags/ as well as refs/heads/
+    # -- and `branches` below is matched against that value.
+    #
+    # An earlier draft of this job was a cron, on the grounds that a tag trigger
+    # races the images: rc-release-pipeline.yml writes rc_<ts>_<sha> as its FIRST
+    # step, while the candidate's images come from docker-publish-ghcr.yml, a
+    # separate push-to-main workflow with a serialising concurrency group. At the
+    # instant an rc_ tag appears its images may still be queued behind earlier
+    # commits', and hack/resolve-rc-target.sh hard-fails on exactly that state by
+    # design. An rc_ tag fires the event at the moment the race is most likely to
+    # be lost.
+    #
+    # staging_ tags carry no such race, which is why the trigger is on them and
+    # not on rc_. nightly-pipeline.yml pushes one only from step-5, which needs
+    # step-3 -- the full E2E matrix -- to have passed, and step-1 ran
+    # scripts/release/verify_candidate_images.sh before any of that. By the time
+    # a staging_ tag exists, the candidate's images have been verified present in
+    # GHCR and then installed and exercised by the matrix. There is no window
+    # left for resolve-rc-target.sh's image gate to fire.
+    #
+    # What the cron bought and this does not is a second attempt: a tick asks
+    # again tomorrow, an event fires once. A run lost to infrastructure is re-run
+    # from Deck, and otherwise the next promotion measures the next candidate.
+    #
+    # And what it changes about coverage: the cron measured the newest candidate
+    # nightly whether or not anything had been promoted, so most candidates were
+    # never measured and some were measured twice. This measures every promotion
+    # and nothing else -- so it runs exactly as often as nightly-pipeline.yml
+    # promotes, which today is on workflow_dispatch only, that file having no
+    # cron yet. Turning its schedule on is what makes this job nightly.
+    cluster: build-kube-agents
+    # The staging promotion tag's exact shape, mirroring STAGING_TAG_SHAPE_REGEX
+    # in scripts/release/common.sh, and not the bare `staging_` prefix that
+    # staging-redeploy-*.yml triggers on. The prefix is a trigger anyone can push
+    # by hand; a `staging_hotfix` typed at a terminal would otherwise take a
+    # project out of a pool shared with the merge-blocking presubmit for hours.
+    # Only the tag nightly-pipeline.yml composes matches this.
+    branches:
+    - ^staging_[0-9]{10}_[0-9a-f]{7}$
+    # No run_if_changed: the promotion is the event, not a file change. A
+    # candidate that touched only documentation still gets measured, because what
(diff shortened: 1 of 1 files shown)


prow/prowjobs/gke-labs/kube-agents/kube-agents-periodics.yaml
```diff
@@ -0,0 +1,297 @@
...
+        #
+        # `|| true` is where "non-gating" lives, and it is one line in a config
+        # on purpose -- hack/ci-eval-rc.sh preserves the real exit status
+        # rather than swallowing it, precisely so that the decision to ignore
+        # it is visible here instead of buried in the script. The cost is that
+        # this job's status says nothing: a RED candidate, a failed deploy and
+        # a broken driver all report success. The verdict a reader wants is in
+        # the rc-eval-summary.md artifact, which the driver writes on every
+        # path including the failure ones.
+        run_step ./hack/ci-eval-rc.sh || true
```
jayantid: 🔴 **Not merge-blocking, but fix before the first tick.** A candidate whose images are not published yet makes this job green with no verdict and no summary.

Line 273 wraps the driver in `|| true`. Lines 270-272 say the driver writes `rc-eval-summary.md` "on every path including the failure ones". It does not. `hack/ci-eval-rc.sh` exits before the summary on resolve failure (line 195, under `set -e`), dirty tree (211), failed checkout (222) and the marker refusal (233).

Resolve failure is this PR's own headline scenario. At 02:00 UTC the newest `rc_*` tag's images can still be queued in `docker-publish-ghcr.yml`. `resolve-rc-target.sh` exits 1, the driver dies at its first assignment, `|| true` swallows it, and Prow reports success. Artifacts hold an empty `rc-target.env` and nothing else. A resolver broken for a month looks identical to a month of green.

Gate on the plumbing, stay advisory on the verdict:

```bash
run_step ./hack/ci-eval-rc.sh || true
# Advisory on the VERDICT, not the plumbing: the driver writes the summary
# only once it has a candidate deployed, or a deploy failure to report.
# No file means it never got that far, and the job should say so.
[ -f "${ARTIFACTS}/rc-eval-summary.md" ]
```

Deploy failures still write a `NOT RUN` summary and stay green as designed. Fix the comment at 270-272 and the PR body to match.

jayantid (approved):

prow/prowjobs/gke-labs/kube-agents/kube-agents-periodics.yaml
stalhaali: follow up: https://github.com/GoogleCloudPlatform/oss-test-infra/pull/2695

PR #2693, opened 2026-09-08
Update release-package TestGrid configurations for parallelism, testing zones, and testing projects.

nicolenliang:
/cc @bkatyl @akerekes 

This PR updates parallelism configs (parallel_count=25—>20, parallel_stagger=60s—>30s) and increases testing zones to ~14-18 zones to further reduce resource/quota strain.


nicolenliang:
Amended to include a change temporarily reverting test projects to run within the shared resource projects for all but E2, for purposes of tracking the current guest agent rollout.

bkatyl (approved):

PR #2705, opened 2026-09-18
feat(kube-agents): gate staging promotion on the release-candidate eval

stalhaali:
## Summary

`post-kube-agents-eval-rc` currently triggers on the `staging_<ts>_<sha>` tag that also deploys the staging estate, and runs its eval driver under `|| true` so the verdict reaches nobody who can act on it. This switches the trigger to a new `evalcand_<ts>_<sha>` tag and removes the `|| true`. Both changes exist to make the same eval a gate instead of a report: the kube-agents nightly pipeline will push `evalcand_`, wait for this job, and push `staging_` — the deploy trigger — only when the verdict is green.

## Why This Change

The eval cannot gate a deploy it races. `staging_` is the ref `staging-deploy.yml` itself triggers on, so a job hung off that tag can only ever report on a deploy that has already happened: the two consume the same push event, and the deploy reaches the cluster in minutes while this job needs 1.5 to 3.5 hours. No amount of wiring on the kube-agents side fixes that, because the ordering problem is in the trigger.

A second tag, pushed first, puts the eval between the two. `evalcand_` names the same commit and shares the same core, so the candidate is still identifiable end to end, but nothing deploys from it.

The `|| true` is the other half. gke-labs/kube-agents#1024 put it there deliberately — "nothing this lane ships may block a release this week" — and that week is over. Without removing it, a red catalog and a clean run are the same commit status.

## What Changed

- `branches` on the postsubmit: `^staging_[0-9]{10}_[0-9a-f]{7}$` → `^evalcand_[0-9]{10}_[0-9a-f]{7}$`. Still the full shape rather than the bare prefix, mirroring `EVALCAND_TAG_SHAPE_REGEX` in `scripts/release/common.sh`, so a hand-typed `evalcand_hotfix` cannot take a project out of the pool the merge-blocking presubmit shares.
- `run_step ./hack/ci-eval-rc.sh || true` → `run_step ./hack/ci-eval-rc.sh`. With `set -e`, the driver's exit status now becomes the job's status and `finished.json`'s `passed`.
- The `description` annotation no longer says "Advisory … non-gating".
- The comments, which are the bulk of the diff. The block explaining why the trigger is a tag push and not a cron, what `RC_TAG`/`PULL_BASE_REF` are for, and what the post-run summary check catches all had to be rewritten around the new meaning.

Three behavioural notes the comments now carry:

- **The promotion reads the summary artifact, not this job's status.** `hack/ci-eval-rc.sh` writes GREEN, RED or NOT RUN into `artifacts/rc-eval-summary.md`, and the exit status collapses the last two: a red catalog and a Boskos lease that never came both exit non-zero. Only RED should hold a candidate back for good. The status is still worth having right — it is what a human reads in Deck and what lands as a commit status — but it is not the gate.
- **A failed deploy now reds this job**, which is a widening. A build that will not install into a clean leased project is not one to put on staging. It does not conflate a broken lane with a bad release, because that path writes NOT RUN and the waiting pipeline withdraws the nomination rather than rejecting the candidate.
- **The `rc-eval-summary.md`-exists check after the driver stays**, narrowed. It no longer has to catch the driver's machinery failures, which now red the job on their own; what it still catches is the driver's three dormancy and trust gates, which print a skip line and exit 0. All three are unreachable from this job as configured — which is exactly the reason to keep the check, since unreachable-as-configured is a property of the config and the failure it produces is silent.

## Context

Part of a three-pull-request change. This one is the Prow side and is safe to merge alone: nothing pushes an `evalcand_` tag until the kube-agents side lands, so the job simply stops firing in the interval. The `staging_`-triggered behaviour it replaces is not depended on by anything else.

- gke-labs/kube-agents#1024 — the issue this lane was built for, including the "nothing this lane ships may block a release this week" constraint being removed here.
- GoogleCloudPlatform/oss-test-infra#2691 — added this job.
- gke-labs/kube-agents PR 2 of 3 — the `evalcand_` tag family, the helpers and the verdict poller.
- gke-labs/kube-agents PR 3 of 3 — the nightly pipeline wiring that pushes the tag and reads the verdict.

## Testing

- `python3 -c "import yaml; yaml.safe_load(...)"` over the changed file: parses, and the job's `name`, `branches` and `description` read back as intended.
- The file is not prettier-formatted on `master` and is not reformatted here.
- No test in this repository covers this job's config; the behaviour it drives is tested on the kube-agents side (`tests/test_ci_eval_rc.py`, `tests/integration/test_seam_rc_eval_verdict.py` in PR 3).

### Live validation

Not live-tested, and it cannot be from here: firing this job requires pushing a tag to `gke-labs/kube-agents` that matches the new `branches` regex, which is the very thing PR 3 adds the machinery for. The first real run is the first nightly after all three land.

What that leaves unverified, stated plainly: that Prow matches `branches` against a tag push with `refs/tags/` stripped. It is asserted by Prow's own `PushEvent.Branch` handling and by the fact that the current `^staging_…$` regex has been firing this job on tag pushes in production, but this change does not re-prove it.

The kube-agents side's own live validation covers the driver's exit paths against a real installation; see that pull request.

## Self-Review

Adversarial pass, run in a subagent handed the diff range and both repositories and nothing else — it did not write the change and was given no reasoning behind it. What it was pointed at: whether the YAML stays schema-correct; every exit path of `hack/ci-eval-rc.sh` now that `set -e` carries the status, and whether the Boskos lease survives each; whether the post-run summary check is still reachable and its stated reason still true; the blast radius of merging this before the kube-agents side; and — because this diff is ~117 lines of comment for three lines of behaviour — every factual claim the comments make, checked against source. Eight findings.

Fixed:

- **The comment told operators to re-run from Deck, and that does not work.** `poll_rc_eval_verdict.py` returns on the *first finished* build at the commit, and every unsettled outcome finishes the build — so by the time a human sees the failure the pipeline has already withdrawn the nomination and exited. A Deck rerun grades a candidate nobody is waiting on. The poller's own summaries say this; the comment contradicted them. Replaced with the opposite advice and the reason.
- **"The next nightly nominates the same commit again" overstates what withdrawal buys.** `resolve_promotion_candidate.sh` resolves the *newest validated* `rc_` tag, and `rc-scheduler.yml` cuts one every three hours, so the next nightly usually measures a newer commit. Deleting the tag makes the commit *eligible* again; it does not queue it. Corrected in both places it appeared.
- **The 360-minute `timeout` now has a hard consumer ceiling of 330 that the comment did not mention.** The poller cannot wait longer, because a GitHub Actions job may not live past 360 minutes. A build that legitimately uses more than 330 finishes green into an empty room. Raising the Prow timeout past 330 now buys nothing on its own, so the `decoration_config` comment says so and points at the poller.
- **No caveat that nothing pushes `evalcand_` tags yet.** The header asserted the finished state. It now records that the job does not fire until the kube-agents side merges, that this is the safe order, and that `ci-health.yml` filters the warning that would otherwise make a stopped lane visible.
- **"What lands as a commit status" as a justification for getting the status right.** The candidate tags are annotated, so the push event's SHA is the tag object's, and GitHub's statuses API takes commits. I did not verify what crier does with that, so rather than assert either way I removed the claim; Deck alone carries the argument, and it is enough.

Reported, deliberately not changed:

- **`max_concurrency: 1` interacts badly with the poller's 45-minute appear clock.** A job held by the concurrency limit is triggered with no pod, uploads no `started.json`, and so reads as `NEVER_RAN` — the nomination is withdrawn and the job then runs for hours producing a verdict nobody reads. Raising the limit to 2 would fix it by letting the pool absorb the overlap, and I did not: this pool is shared with the merge-blocking presubmit, so widening the draw to fix a case that needs a hand-pushed tag on top of a run already in flight trades a real cost against a rare one. `nightly-pipeline.yml`'s own concurrency group prevents the scheduled path from reaching it. The comment now records the interaction, because the `NEVER_RAN` summary blames the tag shape and that would be the wrong diagnosis.
- **The "status has two values, the run has three outcomes" argument is stated in four places** across two repositories, and a fifth copy has already gone stale. Collapsing the in-file duplication was suggested. I kept it: this is a Prow config in a different repository and cannot link into kube-agents source, and the two passages address different readers — one is the header a person reads before anything else, the other is the rationale attached to the `|| true` line itself. The stale fifth copy is a real finding and is fixed in PR 3.

One finding landed outside this diff: two comments in `scripts/eval_dashboard/collect.py` still say the lane is advisory and that a RED candidate reports SUCCESS, which this change falsifies. Fixed in PR 3, where that file lives.

Not covered: nothing was executed. No `checkconfig` run against the oss-test-infra tree, no live Prow, no `gsutil`. Prow-internal behaviour — tag-push postsubmit triggering, crier's status reporting, whether a `max_concurrency`-held job uploads `started.json` — is reasoned from upstream source rather than observed, and the two findings marked plausible above are exactly the ones a single real run would settle. No search was made for sibling pull requests touching `prow/prowjobs/gke-labs/`.

A docs-drift pass was not run separately on this diff: the only prose here is the comments, which the adversarial pass checked claim by claim against source. The kube-agents side got its own docs-drift pass, and the one finding of that pass that concerned this file — the `NOT WIRED YET` precedent `common.sh` sets — is the fix listed above.

## Risk & Rollout

Small blast radius, and the failure mode is silence rather than damage. Between this merging and the kube-agents side merging, no tag matches the new regex and the job never runs — the same state as a night with no candidate. If the kube-agents side is delayed indefinitely, the release-candidate eval simply stops being measured, which is visible as an empty release table on the eval dashboard.

Revert is the one-line `branches` change and restoring `|| true`.

The one thing to watch after the first real run: this job now reds on a failed deploy, so a Boskos project shortage or a GHCR hiccup produces a red postsubmit where it previously produced a green one. That is intended, and the kube-agents side treats it as "no verdict" and retries rather than as a rejection — but it will change what the postsubmit history looks like.

Generated with the help of Claude Opus 5.


prow/prowjobs/gke-labs/kube-agents/kube-agents-postsubmits.yaml
```diff
@@ -76,13 +132,23 @@ postsubmits:
       # after the eval returns, so a killed run leaves the build log and nothing
       # else. That is an argument for raising this on evidence, not for lowering
       # it.
+      #
+      # RAISING IT PAST 330m BUYS NOTHING ON ITS OWN. This is now the larger of
+      # two coupled ceilings: scripts/release/poll_rc_eval_verdict.py waits 330
+      # minutes and then gives up, because a GitHub Actions job may not live
+      # longer than 360 and the surrounding job needs the remainder to write its
+      # summary. A build that legitimately uses more than 330 minutes finishes
```
jayantid, line 140: 🟡 **Not merge-blocking here, but it should hold gke-labs/kube-agents#1760.** This gate arms against a job that has missed its own budget three nights running.

**What's wrong.** This block treats a build over 330 minutes as a hypothetical. The Prow archive says it is the current norm. The eval catalog grew from 78 to 117 units on Sep 16, still at parallelism 4, and the last three runs were deadline-killed at 360m:

| run | fan-out | eval step | job |
|---|---|---|---|
| Sep 15 | 78 units | 265 min | 272 min, passed |
| Sep 16 | 114 units | 359 min | killed at 360m |
| Sep 17 | 117 units | 358 min | killed at 360m |
| Sep 18 | 117 units | 348 min | killed at 360m |

**What happens if it merges.** Nothing here, since nothing pushes `evalcand_` yet. Once #1760 lands, every nightly nomination reaches the poller's 330-minute deadline, is withdrawn, and staging never moves again. The PR description's "1.5 to 3.5 hours" is off by two.

**The fix.** Get the eval under 330 minutes with margin before #1760 merges, then replace the "first real promotions are what should move it" line above with the measured numbers:

```yaml
      # Measured: 78 units ran in 265m (Sep 15); 114-117 units ran 348-359m
      # (Sep 16-18) and were killed at this deadline. The poller gives up at 330m.
```

jayantid (approved):

jayantid (approved):

prow/prowjobs/gke-labs/kube-agents/kube-agents-postsubmits.yaml
stalhaali: Confirmed, and it has got worse since you wrote. Re-read the archive today: Sep 19 and Sep 20 were also killed, so it is five consecutive nights, Sep 16 through Sep 20, all landing at 363–365m against the 365m wall. No successful run since Sep 15.

I have held gke-labs/kube-agents#1760 on this and tracked it as gke-labs/kube-agents#1842.

Two things your table led me to that I think change the fix:

The RC job leaves `EVAL_TASK_PARALLELISM` unset, so it runs the 117-unit catalog at the `hack/ci-eval-pr.sh` default of 4, while `ci-kube-agents-eval-nightly` runs the same catalog at 6. The `decoration_config` comment reasons explicitly about copying the nightly's timeout and the parallelism was not copied with it, which is most of why the cliff hit this job harder than the periodic.

But matching it would not be enough, which is the part I would not have guessed. The nightly at parallelism 6 ran 357.3 / 401.3 / 482.7 / 480.8 / 382.4m over Sep 17–21, overrunning its own 480m budget twice. So the catalog does not clear 330m at 6 either, and "raise the RC job to 6" is a real improvement that still leaves the gate unsatisfiable.

On updating the comment with measured numbers — agreed, and I would rather do it once the fix is chosen than pin today's figures into a block that is about to be wrong again. If the answer turns out to be a parallelism the job sets itself, the comment wants to say that rather than record what 4 cost.

Thanks for catching this before it merged. The failure mode it would have produced is the quiet kind: nomination withdrawn at 330m, staging frozen, nothing red anywhere to point at.

